PR #1, opened 2013-03-25
Resolve "buffer too small" error with some bigger certs

madprogrammer:
This patch allows to use PPTP VPN from Linux to Windows Server using Aladdin eToken PRO 64K. Without the patch, for some certificates pppd will fail with error: "pkcs11_init_cert: buffer too small" and refuse to establish connection.

This fixes ticket #350 in libp11 trac:
https://www.opensc-project.org/opensc/ticket/350


nmav:
This should be already resolved.

PR #3, opened 2013-08-15
Get public exponent from public key, if private key doesn't contain it.

datacompboy:
Some smart cards doesn't save public exponent in private key object; but its required to correctly export public key or create certificate req; this fixes that.

Yes, that doesn't fixes PKCS11_get_key_exponent -- but it shouldn't be used, since it "generic", not RSA


nmav:
Thanks, merged.

PR #10, opened 2014-07-28
fix PKCS11_store_certificate() return -1 when called just after PKCS_CTX...

yusukemihara:
..._load().

in pkcs11_init_cert() (p11_cert.c:156), tpriv->certs = -1 when just after PKCS_CTX_load(),so OPENSSL_realloc(alloc_size = 0) return null.It caused early return.


nmav:
Hi, I'm planning for a release of libp11. Would you like to simplify and squash your patches against master?

nmav:
Closing as too old. Please reopen when updated.

PR #11, opened 2014-09-22
Allow RSA_NO_PADDING padding mode in PKCS11_private_encrypt.

sadt:
Permitting raw RSA operations is usefull when the software
(e.g. openssl) wants to do the padding itself.

### DIFF
--- a/src/p11_ops.c
+++ b/src/p11_ops.c
@@ -97,25 +97,33 @@ PKCS11_private_encrypt(int flen, const unsigned char *from, unsigned char *to,
 	if (key == NULL)
 		return -1;
 
-	if (padding != RSA_PKCS1_PADDING) {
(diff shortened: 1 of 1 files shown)


nmav:
Hello @sadt could you enhance the test suite to check for both signing operations?

sadt:
Hi @nmav,
I'm wondering what you call the "test suite". Do you want me to add a new pgm in the examples directory (say rawsign.c) that calls PKCS11_private_encrypt with RSA_NO_PADDING and tests the result ?

nmav:
That's the idea. You can put it in tests and use a wrapper script on top (see testpkcs11.softhsm) so that is automatically run using softhsm on "make check".

sadt:
I've just made a new pull request #31 which includes a test pgm rawrsasign.c in addition to the initial enhancement code.

PR #18, opened 2015-06-08
Fix two memory leaks with init_args

tiran:
PKCS11_CTX_init_args() duplicates init_args string with strdup(). The
result must be freed with free() eventually. PKCS11_CTX_init_args()
now calls free() on the previously allocated string and PKCS11_CTX_free()
frees priv->init_args, too.


nmav:
Merged.

PR #19, opened 2015-06-08
Add PKCS11_CTX_init_flags()

tiran:
NSS softoken needs CKF_OS_LOCKING_OK. The flag for C_Initialize() can
now be set with Add PKCS11_CTX_init_flags().

Basically I'm trying to achieve the same goal as these two:
http://opensc-devel.opensc-project.narkive.com/b7eqECsN/using-pkcs11-helper-to-load-firefox-pkcs11-provider
https://www.mail-archive.com/opensc-devel@lists.opensc-project.org/msg09054.html


nmav:
Should be handled by #29. Please reopen if not.

PR #21, opened 2015-06-26
Transparent handling of PKCS #11 reinitialization after fork

nmav:
This patch set adds support for transparent call of C_Initialize and reopen of objects after a fork. It uses mutexes to protect the checks.

Unfortunately there is no testsuite to run. I've tested it using nginx and the example code in
https://github.com/OpenSC/engine_pkcs11/issues/16

Please review and test if possible.

### DIFF
--- a/src/p11_ec.c
+++ b/src/p11_ec.c
@@ -120,12 +120,13 @@ static int pkcs11_get_ec_private(PKCS11_KEY * key, EVP_PKEY * pk)
 		prkey = PKCS11_find_key_from_key(key);
 	}
 
-
-	if (!(ec = EVP_PKEY_get1_EC_KEY(pk))) {
-		ERR_clear_error();	/* the above flags an error */
+	if (pk->type == EVP_PKEY_EC) {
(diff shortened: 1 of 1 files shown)


LudovicRousseau:
I am not sure pthread_atfork() should be used in a library. I had an issue in libpcsclite. You cannot *un-register* a handler when your library is unloaded.

Scenario is:
- application A loads libp11
- libp11 uses ``pthread_atfork()`` to register a handler
- application A uses PKCS#11 with no problem
- application A unloads libp11
- application A calls fork()
 - handler is called but the code is no more present in memory => **crash**

See also https://github.com/LudovicRousseau/PCSC/commit/b4d935a73e84b899dbf63bc97bca0c50c9b84f5b

nmav:
Indeed, there even an issue open at the posix group. http://austingroupbugs.net/view.php?id=851
It would then make sense to keep __register_atfork and the getpid() fallback.

nmav:
Patch updated to remove support for pthread_atfork()

PR #24, opened 2015-07-08
Further increase the maximum size of a certificate

nmav:


dengert:
The code should not have a fixed size for a cert. 
    If I read the code correctly, it should make two calls to pkcs11_getattr_var, 
      the first with data = NULL  this would then call PKCS#11
       C_GetAttributeValue  to request the size be set.
      The buffer could then be allocated, and a second call to pkcs11_getattr_var to get the
        cert. 
        
        The problem may be in how pkcs11_getattr_var is
          written., but the intent of PKCS#11 is to allow the caller
          request the size of anything to allocate a buffer to hold it.
          
          
        PKCS#11 2.20 page 133:
       C_GetAttributeValue obtains the value of one or more attributes
      of an object
      3. Otherwise, if the pValue field has the value NULL_PTR, then the
      ulValueLen field is
      modified to hold the exact length of the specified attribute for
      the object
      
      Can you rewrite this code to request the length then allocate the
      correct size buffer?
    
    On 7/8/2015 9:21 AM, Nikos
      Mavrogiannopoulos wrote:
    
    
      
      You can view, comment on, or merge this pull request online
        at:
        https://github.com/OpenSC/libp11/pull/24
      Commit Summary
      
        Further increase the maximum size of a certificate
      
      File Changes
      
         M src/p11_cert.c
          (18) 
      
      Patch Links:
      
        https://github.com/OpenSC/libp11/pull/24.patch
        https://github.com/OpenSC/libp11/pull/24.diff
      
      —
        Reply to this email directly or view it on
          GitHub.
      
        
          
          
        
        
      
    
    
    -- 

 Douglas E. Engert  <DEEngert@gmail.com>

nmav:
I'm not confident with large changes on this library because there is no test suite and requires too much manual testing. Once there is a sufficient test suite I could make larger changes.

dengert:
How about this as an example:
     https://github.com/dengert/libp11/commit/6a8976f207cb6027acc02c43ae7fd82d4f79dcef
    
    (This is untested, can you try it?)
    
    It gets the size first, then allocates the buffer for the correct
    size. 
    
    On 7/9/2015 4:11 AM, Nikos
      Mavrogiannopoulos wrote:
    
    
      I'm not confident with large changes on this library because
        there is no test suite and requires too much manual testing.
        Once there is a sufficient test suite I could make larger
        changes.
      —
        Reply to this email directly or view
          it on GitHub.
      
        
          
          
        
        
      
    
    
    -- 

 Douglas E. Engert  <DEEngert@gmail.com>

nmav:
That is exactly my issue, it is easy to change the code but difficult to test. I'd prefer for pull/22 to be accepted, and then with any change to require a test to verify that the change works.

nmav:
I close this PR since you have a better patch.

dengert:
I have pushed the patch. 
https://github.com/OpenSC/libp11/pull/26

PR #25, opened 2015-07-09
Transparent handling of PKCS #11 reinitialization after fork

nmav:
This patch set adds support for transparent call of C_Initialize and reopen of objects after a fork. It uses mutexes to protect the checks.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -141,7 +141,15 @@ if test "${WIN32}" != "yes"; then
 	)
 fi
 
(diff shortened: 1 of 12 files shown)


nmav:
Checking it further it seems that this patch is incomplete. There is PKCS11_enumerate_keys() and the object handle which is not reopened. I'll do a follow up commit to fix these. Most likely I'll remove any pthread locks as this library is clearly non-thread safe.

PR #29, opened 2015-08-13
Tell PKCS#11 modules that using OS locking primitives is OK

gerow:
This is related to https://github.com/OpenSC/libp11/pull/19 but I feel like there shouldn't really be a need to set init flags if they're only used to tell the library whether or not to use OS locking or OS threading. Just telling the module that using the OS's locking primitives unconditionally should be fine.


nmav:
Merged.

PR #36, opened 2015-11-19
fixed MSVC build and NULL pin handling

mtrojnar:


mtrojnar:
Replaced with https://github.com/OpenSC/libp11/pull/37 and https://github.com/OpenSC/libp11/pull/38.

PR #38, opened 2015-12-01
Fixed prev_pin handling

mtrojnar:
The original PKCS11_login() implementation allows for pin to be NULL. Commit https://github.com/OpenSC/libp11/commit/ecb83dad73ae39afb24b3cea294d44b7f3e3a6f1 breaks it by introducing prev_pin defined as char[64]. This pull request fixes this regression by defining prev_pin as char *.  It also eliminates the use of snprintf() function, which cannot be resolved under MSVC.

### DIFF
--- a/src/libp11-int.h
+++ b/src/libp11-int.h
@@ -66,7 +66,7 @@ typedef struct pkcs11_slot_private {
 	int prev_rw; /* the rw status the session was open */
 
 	/* options used in last PKCS11_login */
-	char prev_pin[64];
+	char *prev_pin;
 	int prev_so;
 } PKCS11_SLOT_private;
 #define PRIVSLOT(slot)		((PKCS11_SLOT_private *) (slot->_private))
--- a/src/p11_slot.c
+++ b/src/p11_slot.c
@@ -228,8 +228,13 @@ int pkcs11_login(PKCS11_SLOT * slot, int so, const char *pin, int relogin)
 		CRYPTOKI_checkerr(PKCS11_F_PKCS11_LOGIN, rv);
 	priv->loggedIn = 1;
 
-	if (priv->prev_pin != pin)
-		snprintf(priv->prev_pin, sizeof(priv->prev_pin), "%s", pin);
+	if (priv->prev_pin != pin) {
+		if (priv->prev_pin) {
+			OPENSSL_cleanse(priv->prev_pin, strlen(priv->prev_pin));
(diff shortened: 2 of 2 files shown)


dengert:
There may still be problems. 

PKCS #11 V2.20  uses  CK_UTF8CHAR_PTR  pPin; CK_ULONG ulPinLen; 
Modified UTF-8 does not allow a 0x00 character, but PKCS#11 calls for UTF8.

 So the PIN could have a NULL character and need not be NULL terminated.  It also says:
"To log into a token with a protected authentication path, the pPin parameter to C_Login should be NULL_PTR."

mtrojnar:
@dengert: Cryptoki (PKCS#11) indeed uses ulPinLen to indicate pin length and not C-style null-terminated strings.  Unfortunately, there is no way to specify the pin length with the libp11 API:
```c
extern int PKCS11_login(PKCS11_SLOT * slot, int so, const char *pin);
```
I agree this is a problem, but unfortunately this is not something that could be addressed without changing the current libp11 API.

I think you may have confused NULL pin value (which is what this pull request fixes) with null-terminated pin value (which is the representation used internally by libp11).

The corresponding engine_pkcs11 code is:
```c
        /* Now login in with the (possibly NULL) pin */
        if (PKCS11_login(slot, 0, pin)) {
            /* Login failed, so free the PIN if present */
            zero_pin();
            fail0("Login failed\n");
        }
```

dengert:
From the PKCS11 API there can be 3 states for the PIN.  The pointer is NULL indicating to try alternate login method. The length of the pin can be zero length indicating that there is no PIN. Or there is a pin of some length. 

Before the fork modifications, the libp11 could do all three as long as the pin is NULL terminated string (which is a subset of UTF8)  But the fork code lost the ability to tell the difference between  a NULL pointer and a zero length string.  

It looks like your mod fixes the regression.

mtrojnar:
This is correct.

In fact, if someone called PKCS11_login() with a non-NULL pin value, and then again with a NULL pin value, the current implementation would use the NULL value as a parameter of snprintf(): https://github.com/OpenSC/libp11/commit/ecb83dad73ae39afb24b3cea294d44b7f3e3a6f1#diff-fd0489f4ed3fb35ba5f49d2d60355d00R191

Also, if someone executed PKCS11_login() with different pin values stored in the same buffer (so the pointer doesn't change), priv->prev_pin would only be updated with the **first** provided pin.

Issue #39, opened 2015-12-02
atfork handler breaks engine initialization


mtrojnar:
Commit https://github.com/OpenSC/libp11/commit/ecb83dad73ae39afb24b3cea294d44b7f3e3a6f1 breaks applications that fork after engine initialization.  Unless "foreground = yes" is specified in stunnel.conf, stunnel returns the following errors:
```
[0]: error queue: 14179006: error:14179006:SSL routines:tls_construct_server_key_exchange:EVP lib
[0]: SSL_accept: 80004003: error:80004003:PKCS11 library:PKCS11_open_session:Invalid slot ID
[1]: error queue: 14179006: error:14179006:SSL routines:tls_construct_server_key_exchange:EVP lib
[1]: error queue: 800280B3: error:800280B3:PKCS11 library:PKCS11_get_attribute:Session handle invalid
[1]: SSL_accept: 800060B3: error:800060B3:PKCS11 library:PKCS11_enum_keys:Session handle invalid
[2]: error queue: 14179006: error:14179006:SSL routines:tls_construct_server_key_exchange:EVP lib
[2]: SSL_accept: 800280B3: error:800280B3:PKCS11 library:PKCS11_get_attribute:Session handle invalid
[3]: error queue: 14179006: error:14179006:SSL routines:tls_construct_server_key_exchange:EVP lib
[3]: SSL_accept: 800280B3: error:800280B3:PKCS11 library:PKCS11_get_attribute:Session handle invalid
...
```
Commenting out the attempted context reloading in the PKCS11_CTX_reload() function fixes the problem.

nmav:
Actually without that patch applications fail. Applications that fork must reinitialize PKCS#11. Could it be that you are using a module that cannot cope with reinitialization?

mtrojnar:
I tested it with with the OpenSC library with SmartCard-HSM and ePass2003 hardware.
@nmav: Which software/hardware did you test?  Which applications failed without this https://github.com/OpenSC/libp11/commit/ecb83dad73ae39afb24b3cea294d44b7f3e3a6f1?

nmav:
opensc was buggy when handling fork. Could it be that you are seeing the issue solved in: https://github.com/OpenSC/OpenSC/pull/494 ?

About changing to explicit fork handling it doesn't work. No application was using it because that's an oddball PKCS#11 requirement no-one knows about. The pkcs11 library should handle it.

mtrojnar:
[...]

PR #43, opened 2015-12-15
Atfork reinitialization fixed to work with opensc-pkcs11 (and not only SoftHSM)

mtrojnar:
According to the PKCS#11 standard:

> C_Initialize should be the first Cryptoki call made by an application, except for calls to
> C_GetFunctionList. What this function actually does is implementation-dependent;
> typically, it might cause Cryptoki to initialize its internal memory buffers, or any other
> resources it requires.

One of the initialized (invalidated) internal buffers mentioned by the standard is the internal slot table (the list of available slots).  This pull request adds missing reinitialization of the internal slot table PKCS11_CTX_reload().

This change fixes https://github.com/OpenSC/libp11/issues/39.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ compiler:
   - gcc
   - clang
 before_script:
+  - sudo apt-add-repository -y ppa:pkg-opendnssec/ppa
   - sudo apt-get update -qq
   - sudo apt-get install -y softhsm libsofthsm-dev opensc
   - touch config.rpath && autoreconf -fvi && ./configure
--- a/src/libp11-int.h
+++ b/src/libp11-int.h
@@ -178,6 +178,7 @@ typedef struct pkcs11_cert_private {
 #define PKCS11_DUP(s) \
 	pkcs11_strdup((char *) s, sizeof(s))
 
+extern int pkcs11_enumerate_slots(PKCS11_CTX *, PKCS11_SLOT **, unsigned int *);
 extern void pkcs11_release_slot(PKCS11_CTX *, PKCS11_SLOT *slot);
 
 extern void pkcs11_destroy_keys(PKCS11_TOKEN *);
--- a/src/p11_attr.c
+++ b/src/p11_attr.c
@@ -42,8 +42,6 @@ pkcs11_getattr_int(PKCS11_CTX * ctx, CK_SESSION_HANDLE session,
 	CK_ATTRIBUTE templ;
 	int rv;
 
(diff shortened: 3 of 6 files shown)


nmav:
This pull requests breaks CI.

mtrojnar:
There seems to be a problem with the atexit handler installed on the CI machine by libsofthsm.so.

I could not reproduce it on Ubuntu 15.10.  I will try again with an older version of libsofthsm.

In the meantime CI has collected (https://travis-ci.org/OpenSC/libp11/jobs/97731482) the following crash:
```
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(__fortify_fail+0x37)[0x2ae086fa5e57]
/lib/x86_64-linux-gnu/libc.so.6(__fortify_fail+0x0)[0x2ae086fa5e20]
/usr/lib/libsofthsm.so(C_DigestFinal+0x132)[0x2ae087680042]
```

mtrojnar:
This pull request somehow triggered a bug in the libsofthsm package included in Ubuntu 12.04.5 LTS used by CI.  The current version of this package works just fine.

I rebuilt the libsofthsm package from Ubuntu 12 with debugging symbols.  The crash details are:
```
Core was generated by `./fork-test /usr/lib/libsofthsm.so 1234'.
Program terminated with signal 11, Segmentation fault.
#0  0x00002b654595000b in C_DigestFinal (hSession=<optimized out>, pDigest=0x2b6545524f40 "", 
    pulDigestLen=0x2b6545523688) at main.cpp:1498
1498	    session->digestPipe->read(pDigest, session->digestSize);
(gdb) print session->digestPipe
$1 = (Botan::Pipe *) 0x0
(gdb) list
1493	
1494	  try {
1495	    session->digestPipe->end_msg();
1496	
1497	    // Returns the result
1498	    session->digestPipe->read(pDigest, session->digestSize);
1499	    *pulDigestLen = session->digestSize;
1500	  }
1501	  catch(std::exception& e) {
1502	    char errorMsg[1024];
```
I will try to update .travis.yml to install a recent version libsofthsm.  It currently uses a version that is over 4 years old.  I guess the relevant SoftHSM NEWS file entry may be:
```
SoftHSM 1.3.2 - 2012-03-07

Bugfixes:
* Fix the destruction order of the Singleton objects.
```

mtrojnar:
@nmav: CI is now fixed.

mtrojnar:
I rearranged the commits so that each commit addresses a separate issue.

PR #44, opened 2015-12-22
Thread-safety for PKCS11_ecdsa_sign, PKCS11_private_encrypt and PKCS11_private_decrypt

mtrojnar:
This pull request allows engine_pkcs11 to be used with multi-threaded applications.

Functions PKCS11_ecdsa_sign, PKCS11_private_encrypt and PKCS11_private_decrypt translate into two separate PKCS#11 operations: initialization and collecting the results.  Most engines only allow for one of those operations to be performed at any given time.  If a libp11 function is preempted in between the initialization and collecting the results, the preempting call fails with the CKR_OPERATION_ACTIVE error.

This pull request implements a conservative solution to the stated problem: the sensitive pairs of operations are wrapped in per-slot critical sections.  This may (unlikely) lead to suboptimal performance if a specific engine can perform more than one concurrent operation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,15 @@ Build state
 libp11 README -- Information for developers
 ===========================================
 
-This is a higher than PKCS #11 library to access PKCS #11 objects.
-It is designed to integrate with applications that use openssl. Note, however,
-that the libp11 library is not thread safe.
+This library provides a higher-level (compared to the PKCS#11 library)
+interface to access PKCS#11 objects.  It is designed to integrate with
+applications that use OpenSSL.
+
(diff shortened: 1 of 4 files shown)


nmav:
Thank you. That's a very nice changeset. Would you like to include a test suite for that? Without it, it looks easy to break that functionality inadvertently. @mtrojnar do you want to take over the maintainership of this library?

mtrojnar:
@nmav I'll be glad take over the maintainership of this library.  The codebase certainly needs some loving care (cleanup and bugfixes) I can provide.

Thread-safety testing is always hard, because race-conditions are non-deterministic by definition.  I will think about implementing a test suite for it.

PR #48, opened 2015-12-30
Fixes access to public keys, including ECC keys.

mouse07410:
Fixes PIN-less access to public keys.
Fixes access to ECC keys.
Adds debugging printouts (upon "verbose") to help locating problems.
Makes PKCS11_enumerate_pubkeys() exportable.

### DIFF
--- a/src/libp11.exports
+++ b/src/libp11.exports
@@ -11,6 +11,7 @@ PKCS11_is_logged_in
 PKCS11_login
 PKCS11_logout
 PKCS11_enumerate_keys
+PKCS11_enumerate_pubkeys
 PKCS11_get_key_type
 PKCS11_get_key_size
 PKCS11_get_key_modulus
--- a/src/p11_attr.c
+++ b/src/p11_attr.c
@@ -29,7 +29,11 @@
 #include <assert.h>
 #include <string.h>
 
+#include <stdio.h>
+
 #include "libp11-int.h"
+
+static int verbose = 0;
 
 /*
  * Query pkcs11 attributes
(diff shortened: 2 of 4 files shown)


mtrojnar:
Fixing access to public keys is already developed in the PKCS11_enumerate_public_keys branch.  Could you contribute a test case for this functionality?

Please create separate pull requests for the remaining issues (PIN, ECC).

mouse07410:
> Fixing access to public keys is already developed in the PKCS11_enumerate_public_keys
> branch. Could you contribute a test case for this functionality?

I'm sorry, I'm not a QA type - I know little of that stuff. And I've nothing that isolates this particular functionality. I've had a practical case/problem to solve (openssl couldn't access keys on the token directly, and/or required PIN to access public keys on the token), and I found a solution, probably not too ugly. Must add that I don't have either skills or time to invest much in this project, especially since the actual problems are solved, as far as I and my applications are concerned.

> Please create separate pull requests for the remaining issues (PIN, ECC).

I will try, but it is problematic for me as it requires that I break/destroy my working code in Github that I hoped to access and use on different computers at different locations (as it is **working**). Now compound this difficulty with my lack of fluency in Git/Github, and you'd probably see that it might be easier for you to cherry-pick from the PR I submitted. After all, there are four files - two are less relevant (`libp11.exports` and `p11_attr.c`), one is mostly for public key access in general (`p11_key.c`), and one (`p11_ec.c`) to access *and properly form* ECC public key.

Since I've pretty much used only your own "building blocks" that you must recognize from your private key access routines, it should be straightforward for you to either integrate them as they are (mostly straightforward copies or additions), or to use in the re-write you're planning to do.

mtrojnar:
>>  Please create separate pull requests for the remaining issues (PIN, ECC).

> I will try, but it is problematic for me as it requires that I break/destroy my working code in Github that I hoped to access and use on different computers at different locations (as it is working). Now compound this difficulty with my lack of fluency in Git/Github, and you'd probably see that it might be easier for you to cherry-pick from the PR I submitted.

This is exactly what I'm going to do.  Thank you for contributing your solution.

> Here's my test case for an RSA token (e.g., RSA keys) with debugging output turned on:

Great.  I will try to port them to SoftHSM, so they can be used for engine_pkcs11 CI testing.

PR #61, opened 2016-01-22
EVP_KEY_CTX_new and _free

dengert:
EVP_KEY_CTX_new and EVP_KEY_CTX_free are new in OpenSSL 1.1
Previously they were called EVP_KEY_CTX_create and EVP_KEY_CTX_destroy.
Use older names.


mouse07410:
Yes, we need this.

mtrojnar:
This is already fixed in the master branch.

Issue #66, opened 2016-02-01
Possible deadlock


bphinz:
Hi,

I'm working on a patch to the serf library that adds support for PKCS11 via engine_pkcs11 (see [serf issue 27](https://issues.apache.org/jira/browse/SERF-27)) and after upgrading libp11 from 0.2.8 to 0.3.1 found what appears to be a deadlock in the new thread safety code.  The load_cert_ctrl function from engine_pkcs11 calls PKCS11_enumerate_certs from libp11, which sets a dynlock on the context and then calls pkcs11_find_certs.  pkcs11_find_certs calls PKCS11_open_session, which in turn calls pkcs11_open_session.  At this point, if it's not a relogin, CHECK_SLOT_FORK is called and the process deadlocks because another lock is requested on the context and it just sits and waits for it.  Any chance someone could take a look at this?  I have very little experience with PKCS11 at all, just jumped into this because the company I work for is adopting smart cards for authentication and we need subversion v1.8.

Thanks,
-brian

bphinz:
Duh, looks like a trivial solution...  Just duplicate the check from pkcs11_find_certs ahead of the lock operation in PKCS11_enumerate_certs:

```diff
--- libp11-0.3.1/src/p11_cert.c
+++ libp11-0.3.1/src/p11_cert.c
@@ -38,11 +38,14 @@
	PKCS11_CERT ** certp, unsigned int *countp)
{
	PKCS11_TOKEN_private *tpriv = PRIVTOKEN(token);
+	PKCS11_SLOT *slot = TOKEN2SLOT(token);
	PKCS11_CTX *ctx = TOKEN2CTX(token);
	PKCS11_CTX_private *cpriv = PRIVCTX(ctx);
	int rv;

	if (tpriv->ncerts < 0) {
+		if (!PRIVSLOT(slot)->haveSession && PKCS11_open_session(slot, 0))
+			return -1;
		pkcs11_w_lock(cpriv->lockid);
		rv = pkcs11_find_certs(token);
		pkcs11_w_unlock(cpriv->lockid);
```

I'll submit a PR later this afternoon.

mtrojnar:
Good point.  `PKCS11_open_session()` indeed cannot be invoked inside the critical section.
I improved the fix and committed it as https://github.com/OpenSC/libp11/commit/c730ba61cefd943f40493475391f7253eae28b66.

bphinz:
[...]

PR #67, opened 2016-02-01
Avoid potential deadlock

bphinz:
[re #66] If PKCS11_enumerate_certs is called by a multi-threaded application
and there is not an open session, a deadlock condition can occur
when PKCS11_open_session is called and it is not a relogin case.


mtrojnar:
Good point.  `PKCS11_open_session()` indeed cannot be invoked inside the critical section.
I improved the fix and committed it as https://github.com/OpenSC/libp11/commit/c730ba61cefd943f40493475391f7253eae28b66.

PR #88, opened 2016-08-01
tests: do not assume that public keys are available without PIN

nmav:
SoftHSM by default sets CKA_PRIVATE to true for all objects created, including public keys. Old versions of pkcs11-tool did not explicitly set such objects to have CKA_PRIVATE=false, thus
it may happen that this test will fail on such systems. For that, login early prior to checking the public objects.

### DIFF
--- a/examples/listkeys.c
+++ b/examples/listkeys.c
@@ -77,13 +77,6 @@ int main(int argc, char *argv[])
 	printf("Slot token model.......: %s\n", slot->token->model);
 	printf("Slot token serialnr....: %s\n", slot->token->serialnr);
 
-	/* get public keys */
-	rc = PKCS11_enumerate_public_keys(slot->token, &keys, &nkeys);
-	error_queue("PKCS11_enumerate_public_keys");
-	CHECK_ERR(rc < 0, "PKCS11_enumerate_public_keys failed", 4);
-	CHECK_ERR(nkeys == 0, "No public keys found", 5);
-	list_keys("Public keys", keys, nkeys);
-
 	if (slot->token->loginRequired && argc > 2) {
 		strcpy(password, argv[2]);
 		/* perform pkcs #11 login */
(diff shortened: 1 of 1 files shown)


mouse07410:
Perhaps a better way would be to *not* set `CKA_PRIVATE = true` for public keys? SoftHSM can tell when a public key is created, can't it?

mtrojnar:
One of the main goals of this test is to make sure that public key operations work without PIN (this is a recent feature of libp11).  This PR defeats this goal.

BTW: How old exactly are those old versions of pkcs11-tool?  I think there is more value in properly testing our current functionality than in implementing workarounds for obsolete versions of other packages.

mouse07410:
> How old exactly are those old versions of pkcs11-tool? 

I do *not* see this behavior in the current (0.16.0) version, and am pretty sure it was not present in 0.15.0.

> I think there is more value in properly testing our current functionality than in implementing workarounds for obsolete versions of other packages.

Agree 100%. Especially when those "workarounds" would break what's currently working.

nmav:
Since this test is hard-coded to use softhsm + pkcs11-tool it should take into account that to work properly it requires a specific pkcs11-tool version. Softhsm according to their developers behave as expected by setting CKA_PRIVATE to TRUE by default on its tokens for any object.
https://issues.opendnssec.org/browse/SUPPORT-170

It is only after commit 4df35b922c8eb7e0776a23260b65e570b33e4d42 that pkcs11-tool explicitly marks these objects as non-private.

The best would be to simply identify the version of opensc being installed and skip this test, but opensc doesn't seem to install a pkg-config file or something similar.

nmav:
According to tests with fedora, the issue exists in opensc 0.14.0 and earlier. I've created a PR on opensc to install a pkg-config so that we can detect the version, but for now there is not much that can be addressed, other than test suite failing with older opensc versions.

mouse07410:
> Softhsm according to their developers behave as expected by setting CKA_PRIVATE to TRUE by default on its tokens for any object. https://issues.opendnssec.org/browse/SUPPORT-170

I would bring this issue up with the developers again. Maybe they've seen the light since then. If needed, I'd fork SoftHSM.

PR #93, opened 2016-08-11
Allow override of openssl library name

matthauck:
FIPS enabled builds of openssl call the library "libeaycompat32.lib"
instead of just "libeay32.lib".


mtrojnar:
AFAIK OpenSSL only creates `libeaycompat32.lib` for static fips builds.  The basic purpose of libp11 is to convert PKCS#11 objects into OpenSSL objects (and the other way around), so the application also needs the OpenSSL library.  Static linking of libp11 against OpenSSL and then the application against OpenSSL would result in having two separate OpenSSL libraries linked into the final application.  This would not only produce a very large binary, but even more importantly those two OpenSSL libraries would use two separate global global data segments.  I doubt OpenSSL would reliably work with such configuration.

Please correct me if I missed something.

matthauck:
Correct. We are compiling OpenSSL statically. We are okay with the two separate copies this results in. This may not be the most common configuration, but it is a valid one I believe -- libp11 has been working just fine with this setup.

mtrojnar:
One thing that's almost certainly broken is thread safety (not everyone needs it).  The other one is FIPS, which is never enabled for the OpenSSL instance used by libp11.  I'm not sure how many other functionalities may be affected.

Yes, it may appear to work or even actually work in some very simple configurations.  I don't, however, think we should support such configurations.

matthauck:
Hmm. It seemed to me that this is supported by openssl. It looks like it may be happening in [IMPLEMENT_DYNAMIC_BIND_FN](https://github.com/openssl/openssl/blob/OpenSSL_1_0_1-stable/crypto/engine/engine.h#L827) in engine.h. My read of that is that it detects if the "static state" of the engine is different and copies over the locking callbacks in that case.

matthauck:
FWIW, I used to see eager-initialization of the libp11 library on `ENGINE_init`, and then I turned on dynamic locking and it switched to using delayed initialization due to the checks in [pkcs11_init](https://github.com/OpenSC/libp11/blob/32d739b/src/eng_back.c#L253-L255). This is another indication that thread-safety is actually working properly as I believe it to be.

mtrojnar:
Yes, I indeed forgot about IMPLEMENT_DYNAMIC_BIND_FN.  Good point.

What about FIPS?  FIPS validation requires that __only__ certified cryptographic code is enabled.

mtrojnar:
http://csrc.nist.gov/publications/fips/fips140-2/fips1402.pdf says:
> All cryptographic software and firmware shall be installed in a form that protects the software and firmware source and executable code from unauthorized disclosure and modification.

IANAL, but this requirement does not seem to only affect the cryptographic code that is _exercised_.

Also, libp11 __does__ use cryptographic code from the OpenSSL library for RSA key generation and  for hash functions in some signature mechanisms (I'd need to check if this functionality is used by the engine interface).

matthauck:
Fair point. The distinction would be for the use case that does not use `libp11` functions per se, but only uses the pkcs11 engine library.

mtrojnar:
I think this PR is not really useful, because statically linked OpenSSL would not be FIPS-compliant anyway.
On the other hand the PR is probably quite harmless.  I'd appreciate comments from other FIPS users.

matthauck:
argh. 💢  Turns out there is definitely some broken thread-safety with this static build setup. Didn't realize that libp11.so/dll had its own static copy of openssl, not just pkcs11.so/dll. So even though pkcs11 library gets configured properly with respect to crypto locking callbacks, libp11 does not. 

We have specific reasons why our crypto module needs to link openssl statically. If libp11 were to support a statically linked openssl, it looks like the options would be:
1. Modify libp11 build to add an option to compile pkcs11.dll as a standalone shared library with no dependency on a separate libp11.dll. 
2. Do something like IMPLEMENT_DYNAMIC_BIND_FN to pass callback/memory data into libp11.dll the same way it is getting passed into pkcs11.dll.

mtrojnar:
> We have specific reasons why our crypto module needs to link openssl statically.

Knowing them might help finding a solution.

Both options you described seem feasible.  It may be hard to convince someone to implement them without proving that the code would be useful to the community.

matthauck:
> Knowing them might help finding a solution.

Wasn't trying to be secretive, was just trying to avoid boring you with details or derailing the discussion. We have a crypto module that statically links openssl and which our application dynamically links against. It exposes a tiny subset of what openssl can do, thus limiting the surface area primarily for the simplification of FIPS validation. 

The only reason I can think of that one of these features may be useful to the community is that openssl has support for shared engine modules with a static openssl build. So, I think it would benefit the community to have the flexibility of working with a static openssl build.

mtrojnar:
The first solution seems much simpler.  All you need seems to be a linker invocation with all the object files from both libraries, and the exports from pkcs11.dll.

I would like the pkcs11 engine .so/.dll to include libp11 functionality by default from the next release.  This would significantly simplify bundling the engine with other products.

Any comments?

matthauck:
Awesome -- I think that would suit our needs perfectly!

matthauck:
Closing this PR since using a static FIPS library is not generally useful to the community.

Issue #99, opened 2016-09-05
PKCS11_enumerate_certs() returns cached results after login


dwmw2:
Some certificates are stored in tokens with the CKA_PRIVATE flag, so they aren't visible until you log in.

In OpenConnect I call `PKCS11_enumerate_certs()` to find the cert requested by the user. If no matching cert if found, and if only one token was found that matched the user's specification, then I log in to that token... and call `PKCS11_enumerate_certs()` again. This doesn't work because the second call to `PKCS11_enumerate_certs()` will immediately return a cached empty set of results, without asking the (now logged in) token again.

mouse07410:
> Some certificates are stored in tokens with the CKA_PRIVATE flag, so they aren't visible until you log in.

This is conceptually wrong. What tokens do that, and why the provisioner did not reconfigure this?

The whole idea is that one can use *public* information on the token *without* having to log in and provide the PIN, while *sensitive* operations (signing and decrypting or deriving) require PIN. This is used heavily in the scripts and tests, for example.

dwmw2:
With PKCS#11 as far as I can tell, you can basically assume that for *any* particular offence that is conceptually wrong, technically wrong, or morally wrong, there'll be someone out there who does it, for whom it works in their environment, and who expects it to work :)

For this particular case, it seems to be common enough that NSS considers its own internal token, which *does* reliably make certificates visible, to be a special case. It knows it needs to log in to other tokens before it concludes that a given certificate isn't present.

mouse07410:
Until recently SoftHSMv2 had public keys (and certificates?) CKA_PRIVATE by default - you needed to explicitly specify otherwise if you wanted the "sane" behavior. But a few weeks ago, they changed that to what the standard recommends, so now SoftHSMv2 default is CA_PRIVATE for private keys, and PUBLIC for the non-sensitive stuff.

I suspect that NSS does not care for certificates that are not accompanied by private keys, therefore it wants you to login anyway, so it can ascertain that if a fitting certificate is found - there is a private key with it.

I still don't know what your use case is (i.e., what *tokens* do not show their certificates unless logged in - NSS it not a token :), and what you propose to do to remedy the problem you're facing.

dwmw2:
[...]

PR #100, opened 2016-09-05
Destroy cert cache on login/logout too

dwmw2:
Certificates can have the CKA_PRIVATE attribute, so that you need to log
in before you can see them. So destroy the cache when we log in, just as
we do the cache of keys.

Addresses issue #99. Note the concern expressed there by @mouse07410 that *"re-reading the objects from the token upon login would break the ability to sign (because SIGN operation must be immediately proceeded by PIN VERIFY according to the standard SP 800-73), so that's not an option."*

Although as noted, I don't understand the objection. After all, that situation is precisely what the PKCS#11 attribute `CKA_ALWAYS_AUTHENTICATE` is for, and we already *do* destroy the cached keys (as seen in the context of this patch); just not the certs (yet).

### DIFF
--- a/src/p11_slot.c
+++ b/src/p11_slot.c
@@ -195,6 +195,7 @@ int pkcs11_login(PKCS11_SLOT * slot, int so, const char *pin, int relogin)
 		if (slot->token) {
 			pkcs11_destroy_keys(slot->token, CKO_PRIVATE_KEY);
 			pkcs11_destroy_keys(slot->token, CKO_PUBLIC_KEY);
+			pkcs11_destroy_certs(slot->token);
 		}
 		if (spriv->loggedIn) {
 			/* already logged in, log out first */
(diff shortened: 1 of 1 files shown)


mouse07410:
This patch is not huge. Will the existing software that relies on libp11 automatically reload the certs if the cache is deleted, like it does with public keys?

dwmw2:
I don't believe the cache is externally exposed. `PKCS11_enumerate_certs()` will always ask the token if there's no existing cache, so this should all work as expected.

mouse07410:
Then we need to try it, and if it doesn't break other things, like OpenSSL<-libp11<-OpenSC and Tokend (on Mac OS X) call it a day?

mtrojnar:
Thank you for testing.  The PR looks fairly harmless.  I'll merge it.

Issue #107, opened 2016-09-28
SSL connection fails with 'Not a PKCS#11 key'.


dwmw2:
I imported a test key into my Yubikey. It works fine...

```
[dwoodhou@i7 tool]$ ./yubico-piv-tool -a import-key -i  ~/ca/key.pem  -s 9a
Successfully imported a new private key.
[dwoodhou@i7 tool]$ ./yubico-piv-tool -a import-certificate -i  ~/ca/cert.pem  -s 9a
Successfully imported a new certificate.
[dwoodhou@i7 tool]$ openconnect -c ~/ca/cert.pem -k 'pkcs11:manufacturer=piv_II;id=%01;pin-value=1234' localhost:4443
POST https://localhost:4443/
Connected to 127.0.0.1:4443
Using client certificate '192.168.123.1'
SSL negotiation with localhost
```
But OpenSSL doesn't work...
```
 $ openssl s_client -engine pkcs11 -key 'pkcs11:manufacturer=piv_II;id=%01;pin-value=1234' -keyform engine -cert ~/ca/cert.pem -connect localhost:4443
engine "pkcs11" set.
CONNECTED(00000008)
depth=0 CN = localhost
verify error:num=20:unable to get local issuer certificate
verify return:1
depth=0 CN = localhost
verify error:num=21:unable to verify the first certificate
verify return:1
wtf pkcs11_ecdsa_sign_sig 0x24e9ef0
139888915887992:error:80029407:Vendor defined:func(41):Not a PKCS#11 key:p11_ec.c:257:
139888915887992:error:2D06D075:FIPS routines:fips_pkey_signature_test:test failure:fips_post.c:168:
139888915887992:error:2D08E06B:FIPS routines:FIPS_CHECK_EC:pairwise test failed:ec_key.c:249:
139888915887992:error:1409802B:SSL routines:ssl3_send_client_key_exchange:reason(43):s3_clnt.c:2871:
```
The 'wtf' printf there is added by me. Fedora 24 with OpenSSL 1.0.2, with libp11 rebuilt from today's master.

This isn't an EC key; it's an RSA key. I have no idea what's going on here.

dwmw2:
```
Breakpoint 1, pkcs11_load_private_key (ctx=0x6b4290, s_key_id=0x7fffffffe184 "pkcs11:manufacturer=piv_II;id=%01;pin-value=1234", ui_method=0x6a3540, 
[...]

PR #108, opened 2016-09-28
Do not set default RSA and EC methods

dwmw2:
We have our own RSA and EC methods which we set on any EVP_PKEY we hand
out. But they are not generic methods which can be used with software
keys. Do not register them as such.

https://github.com/OpenSC/libp11/issues/107

### DIFF
--- a/src/eng_front.c
+++ b/src/eng_front.c
@@ -211,22 +211,6 @@ static int bind_helper(ENGINE *e)
 			!ENGINE_set_ctrl_function(e, engine_ctrl) ||
 			!ENGINE_set_cmd_defns(e, engine_cmd_defns) ||
 			!ENGINE_set_name(e, PKCS11_ENGINE_NAME) ||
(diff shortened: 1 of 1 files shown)


dwmw2:
(Don't merge yet; OpenSSL has issues with failing to call ENGINE_init(), and having default methods, even if they're broken, has magically saved us so far. I'll file an OpenSSL PR too and let's get confirmation that this really is the right fix)

mtrojnar:
Shouldn't we delay merging this PR until the fixed OpenSSL becomes widely used?

dwmw2:
Nah. It's a bugfix, and although I've just worked around it in curl it was causing actual breakage there — that's where I first saw it. It's more important to fix the bug so that things work in real applications, than it is to support test invocations via the openssl command line.

If you really want, we could try to contrive a hackish workaround where we actually *do* register a working default algorithm for something, calling through to the real OpenSSL implementation whenever we get asked to perform that operation.

Issue #119, opened 2016-10-05
Current github master fails to pass the correct PIN to the token


mouse07410:
PIN handling in the current libp11 master is broken. For the context-specific PIN the code appears to use random data (is it the case of use-after-free?):
```
$ p11tool --list-tokens
Token 0:
	URL: pkcs11:model=p11-kit-trust;manufacturer=PKCS%2311%20Kit;serial=1;token=Default%20Trust
	Label: Default Trust
	Type: Trust module
	Manufacturer: PKCS#11 Kit
	Model: p11-kit-trust
	Serial: 1
	Module: p11-kit-trust.so


Token 1:
	URL: pkcs11:model=p11-kit-trust;manufacturer=PKCS%2311%20Kit;serial=1;token=System%20Trust
	Label: System Trust
	Type: Trust module
	Manufacturer: PKCS#11 Kit
	Model: p11-kit-trust
	Serial: 1
	Module: p11-kit-trust.so


Token 2:
	URL: pkcs11:model=PKCS%2315%20emulated;manufacturer=piv_II;serial=4ff5d59a2a23a8ad;token=PIV%20Card%20Holder%20pin%20%28PIV_II%29
	Label: PIV Card Holder pin (PIV_II)
	Type: Hardware token
	Manufacturer: piv_II
	Model: PKCS#15 emulated
	Serial: 4ff5d59a2a23a8ad
	Module: /Library/OpenSC/lib/opensc-pkcs11.so


Token 3:
	URL: pkcs11:model=YubiKey%20NEO;manufacturer=Yubico;serial=1234;token=YubiKey%20PIV
	Label: YubiKey PIV
	Type: Hardware token
	Manufacturer: Yubico
	Model: YubiKey NEO
	Serial: 1234
	Module: /usr/local/lib/libykcs11.dylib


$ openssl dgst -engine pkcs11 -keyform engine -sign "pkcs11:manufacturer=piv_II;object=SIGN%20key;type=private" -sha384 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -out t384.dat.sig t384.dat
engine "pkcs11" set.
PKCS#11 token PIN: 
Error Signing Data
140735251185744:error:800090A0:Vendor defined:PKCS11_rsa_encrypt:PIN incorrect:p11_rsa.c:120:
[...]

PR #121, opened 2016-10-07
These changes fix the problems introduced by 7573c6e

mouse07410:
These changes fix the problem with non-working UI for context-specific PIN prompt, and also ensure that C_Login is called only when actually needed, as indicated by the token itself.

Notice that if `C_SignInit()` returns any error besides `CKR_USER_NOT_LOGGED_IN`, trying to do `C_Login()` does not make much sense. And if it returns `OK`, then the immediately preceding operation must have been `C_Login` and there is no reason to re-authenticate now.

### DIFF
--- a/src/p11_ec.c
+++ b/src/p11_ec.c
@@ -209,7 +209,7 @@ static int pkcs11_ecdsa_sign(const unsigned char *msg, unsigned int msg_len,
 	CRYPTO_THREAD_write_lock(PRIVSLOT(slot)->rwlock);
 	rv = CRYPTOKI_call(ctx,
 		C_SignInit(spriv->session, &mechanism, kpriv->object));
-	if (!rv)
+	if (rv == CKR_USER_NOT_LOGGED_IN)
 		rv = pkcs11_authenticate(key);
 	if (!rv)
 		rv = CRYPTOKI_call(ctx,
--- a/src/p11_key.c
+++ b/src/p11_key.c
@@ -347,6 +347,8 @@ int pkcs11_authenticate(PKCS11_KEY *key)
 	UI *ui;
 	int rv;
 
+	memset(pin, 0x00, MAX_PIN_LENGTH);
+	
 	if (!kpriv->always_authenticate)
 		return 0;
 
@@ -363,7 +365,7 @@ int pkcs11_authenticate(PKCS11_KEY *key)
 	ui = UI_new();
 	if (ui == NULL)
 		return PKCS11_UI_FAILED;
-	UI_set_method(ui, kpriv->ui_method);
+	UI_set_method(ui, UI_get_default_method());
 	UI_add_user_data(ui, kpriv->ui_user_data);
 	if (!UI_add_input_string(ui, "PKCS#11 key PIN: ",
 			UI_INPUT_FLAG_DEFAULT_PWD, pin, 1, MAX_PIN_LENGTH)) {
--- a/src/p11_rsa.c
+++ b/src/p11_rsa.c
@@ -99,7 +99,7 @@ int pkcs11_private_encrypt(int flen,
 	/* Try signing first, as applications are more likely to use it */
 	rv = CRYPTOKI_call(ctx,
 		C_SignInit(spriv->session, &mechanism, kpriv->object));
-	if (!rv)
+	if (rv == CKR_USER_NOT_LOGGED_IN)
 		rv = pkcs11_authenticate(key);
 	if (!rv)
 		rv = CRYPTOKI_call(ctx,


dengert:
You said: "And if it returns OK, then the immediately preceding operation must have been C_Login and there is no reason to re-authenticate now."

As noted in #119 it looks like libp11 can not assume it can reuse the UI_method use to get the original PIN. The best it can do is use the UI_get_default_method()

dwmw2:
It could also try the one set with the `SET_USER_INTERFACE` engine cmd.

mtrojnar:
Thank you.  It makes sense.  The PR, however, addresses a few separate issues.  I will split them into separate commits.

mouse07410:
This PR addresses two and a half issues :) :

1. Failure to get the correct/usable UI_method. One-liner fix (tested) for `src/p11_key.c`. 
2. Improper logic of dealing with `CKA_ALWAYS_AUTHENTICATE`. One-liner fixes (tested) to `src/p11_rsa.c` and `src/p11_ec.c`.

2.1. In general it is considered good manners to initialize/zeroize allocated buffers. One-liner addition to `src/p11_key.c` to zeroize `pin[MAX_PIN_LENGTH]` before any use of it.

You also might consider changing the minimal allowed PIN length in `src/p11_key.c` line 372 from 1 to, say, 4. 

> I will split them into separate commits.

I understand that you are going to do that, i.e., you don't expect anything from me at this point. Correct?

dengert:
In your testing do you see the prompt: "PKCS\#11 key PIN: " and/or the prompt: "PKCS\#11 token PIN: "?

Does C_SignInit  ever return CKR_USER_NOT_LOGGED_IN? Was this because there never was a C_Login for the initial PIN, or because C_SignInit is is attempting to signal that the key has the CKA_ALWAYS_AUTHENTICATE=TRUE.  I don't see any PKCS#11 requirement for C_SignInit to return CKR_USER_NOT_LOGGED_IN just because the key has the CKA_ALWAYS_AUTHENTICATE=TRUE. 

PKCS#11 2.40: says:
"Re-authentication occurs by calling C_Login with userType set to CKU_CONTEXT_SPECIFIC
immediately after a cryptographic operation using the key has been initiated (e.g. after C_SignInit)."

" Failing or omitting to reauthenticate when CKA_ALWAYS_AUTHENTICATE is set to CK_TRUE will result in CKR_USER_NOT_LOGGED_IN to be returned from calls using the key."

 C_SignInit could return 0. (the above say failing or omitting  to reauthenticate but the reauthenticate is  done after the C_SignInit.

I ask, all this because your change may be working because the PKCS\#11 implementation's C_SignInit is returning  CKR_USER_NOT_LOGGED_IN because of the CKA_ALWAYS_AUTHENTICATE. 

Or it may be working because the underlying PKCS\#11 implementation is not handling CKA_ALWAYS_AUTHENTICATE correctly or it is caching the PIN. 

I think the
` if (rv == CKR_USER_NOT_LOGGED_IN)` 
should be at least:
`if (rv == 0 || rv == CKR_USER_NOT_LOGGED_IN)`

pkcs11_authentication test if the key has CKA_ALWAYS_AUTHENTICATE and returns 0 if it does not.

mouse07410:
I'm happy with the changes applied to the master.

mouse07410:
> In your testing do you see the prompt: "PKCS#11 key PIN: " and/or the prompt: "PKCS#11 token PIN: "?

I did see both. Not any more - because now the redundant re-auth is avoided when it isn't needed.

Page 109 of [pkcs11-base-v2.40-os.pdf](http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.pdf) states that `C_SignInit()` can return `CKR_USER_NOT_LOGGED_IN`. 

Page 51 of that document explains that in that case  there is no need to repeat `C_SignInit()`. One should then perform `C_Login()` and proceed with `C_Sign()`. 

> Does C_SignInit ever return CKR_USER_NOT_LOGGED_IN?

Yes it does - see page 109 of [pkcs11-base-v2.40-os.pdf](http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.pdf).

> Was this because there never was a C_Login for the initial PIN, or because C_SignInit is is attempting to signal that the key has the CKA_ALWAYS_AUTHENTICATE=TRUE

Who cares? It simply says that when (re-)authentication is required, that error code shall be returned (page 51 that you yourself quoted a few times from). And the valid response from the application is `C_Login()`.

> I don't see any PKCS#11 requirement for C_SignInit to return CKR_USER_NOT_LOGGED_IN just because the key has the CKA_ALWAYS_AUTHENTICATE=TRUE

You quoted that requirement yourself a few lines later:
> " Failing or omitting to reauthenticate when CKA_ALWAYS_AUTHENTICATE is set to CK_TRUE will result in CKR_USER_NOT_LOGGED_IN to be returned from calls using the key."

What is the above if not a requirement for the card to return error if authentication did not precede the operation on the CKA_ALWAYS_AUTHENTICATE private key? Is `C_SignInit()`  a "call using the key" (check the format of this call on pages 108-109, and observe the call argument `CK_OBJECT_HANDLE hKey` to be really sure)? Does it not fall under the above clause?

> `C_SignInit()` could return 0. 

That means the authentication immediately preceded `C_SignInit()`, and no repetition is necessary. 

Or it could return `CKR_USER_NOT_LOGGED_IN`, which would mean "the token may be logged in, but another operation was performed after that, so because of `CKA_ALWAYS_AUTHENTICATE`  re-auth is required - I'm expecting `C_Login()` now".

> (the above say failing or omitting to reauthenticate but the reauthenticate is done after the C_SignInit.

Yes, if and when required - as signaled by the return code from `C_SignInit()`. Failing or omitting to (re)authenticate at the later stages of the operation (like, after `C_Sign()` failed with `CKR_USER_NOT_LOGGED_IN`) would not be recoverable, AFAICT. The way I read the standard, you won't be able to get this error from `C_Sign()` unless you ignored it at `C_SignInit()` (in other words, situation when `C_SignInit()` returns `CKR_OK`, but following `C_Sign()` returns `CKR_USER_NOT_LOGGED_IN` is impossible).


> Or it may be working because the underlying PKCS#11 implementation is not handling `CKA_ALWAYS_AUTHENTICATE` correctly 

It is not the job or responsibility of a PKCS#11 implementation to handle `CKA_ALWAYS_AUTHENTICATE`, correctly or otherwise. It is the card's responsibility.

`CKA_ALWAYS_AUTHENTICATE` is *not* an instruction for the application talking to a card - it is an instruction for the *card* to refuse a private key operation unless it was preceded by authentication. Which makes perfect sense from both security and usability points of view.

The only way an application can cheat here is by caching the PIN and issuing `C_Login()` by itself without prompting the user.

It that aspect ("handled by card" vs. "handled by app") it is similar to `CKA_PRIVATE` - try to cheat and write a PKCS#11 implementation that decides to "not handle correctly" `CKA_PRIVATE` and access a private key without logging in first.

> or it is caching the PIN.

No it does not - check the source code to make sure, and check the PKCS11SPY traces I provided. And/or run it yourself with PKCS11SPY and observe the results.

> I think the `if (rv == CKR_USER_NOT_LOGGED_IN)` should be at least `if (rv == 0 || rv == CKR_USER_NOT_LOGGED_IN)`

I disagree. The above is exactly what we are (successfully) trying to avoid. Don't bother user asking for something that the card does not want at this time.

PR #131, opened 2016-10-26
Tie the libp11 soname with openssl soname

### LINKED ISSUE #105: Parallel install with OpenSSL 1.0.2 and 1.1
Since OpenSSL 1.1 introduces a lot of API changes, it's likely to take applications a while to update. At least Fedora is planning to ship an openssl102 compat package in parallel with OpenSSL 1.1:
https://fedoraproject.org/wiki/Changes/OpenSSL110

This is fun because Fedora [packaging guidelines](https://fedoraproject.org/wiki/Packaging:SSLCertificateHandling) say that any package which accepts cert files should also accept RFC7512 PKCS#11 URIs by default, and we're filing bugs against any package that doesn't.

This means we need at least the engine to be built and installed in parallel for both OpenSSL 1.0.2 and 1.1. And maybe libp11 itself too.

To start with, it would be useful if libp11 actually used pkg-config to file the version of OpenSSL that it's building with. We don't seem to pick up the correct cflags/ldflags for OpenSSL at all, and fixing that would allow it to be built against a specific version of OpenSSL.

nmav:
This ensures that no applications will be linked with and old openssl version and new libp11 and vice-versa (also indirectly via the versioning, that no symbol clashes will occur even if both versions of libp11 are in a process). In practice this makes libp11.so.2 when linked with openssl 1.0.x, and libp11.so.3 when linked with 1.1.x.

Resolves #105

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2,6 +2,7 @@ dnl -*- mode: m4; -*-
 
 AC_PREREQ(2.60)
 
+# When bumping versions see also the LT vesion numbers below.
 define([PACKAGE_VERSION_MAJOR], [0])
 define([PACKAGE_VERSION_MINOR], [4])
 define([PACKAGE_VERSION_FIX], [3])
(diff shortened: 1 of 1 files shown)


nmav:
Right. I have updated the text, but I'm not sure if it is sufficient. That change will indeed add quite some complication if that library decides to break ABI at some point. I've mentioned in that text, that in that case either we should support only a single openssl ABI (i.e., only 1.1.0), or bump both in a way that they do not clash.

nmav:
I'm not sure whether following the openssl ABI is the best solution to address the problem of an app linking to openssl 1.0.2, and libp11 (which is linked with 1.1.0) or vice versa. However, I am not aware of a better solution.

PR #133, opened 2016-10-26
save pincode

AndreiKorikov:
for use pkcs#11 url with "pin-value=1234567890"


mtrojnar:
Good point.  I introduced this bug in https://github.com/OpenSC/libp11/commit/32feafd257728c0edf4e6e016e0be06c1376e547#diff-10fed657d6304e172d83db709017dec7R198.  I will merge this PR later today.

PR #134, opened 2016-10-29
build: fix symlink installation

alonbl:
this fixes incorrect fix b313f00cc of symlink installation
by adding dependency instead of overwriting automake behavior.

Signed-off-by: Alon Bar-Lev <alon.barlev@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -123,21 +123,21 @@ AC_ARG_WITH(
 AC_ARG_WITH(
 	[enginesdir],
 	[AS_HELP_STRING([--with-enginesdir], [OpenSSL engines directory])],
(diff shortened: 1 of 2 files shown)


zedinosaur:
This should work, but somewhat annoyingly changes the flag name. You should include a note in the NEWS file at least.

alonbl:
On 29 October 2016 at 22:20, Eric Dorland <notifications@github.com> wrote:

> This should work, but somewhat annoyingly changes the flag name. You
> should include a note in the NEWS file at least.
>

It does not change the flag name as far as I can understand.

zedinosaur:
Your right I misread the patch, thanks!

Issue #122, opened 2016-10-08
Pkcs11 engine causes crash in OpenSSL 1.1.0b in RSA_public_decrypt.


nased0:
Hello!
I have compiled libp11 in VS2015 for OpenSSL 1.1.0b by modifying OPENSSL_LIB in make.rules.mak .
Unfortunately engine pkcs11.dll crashes OpenSSL 1.1, because it does not set RSA method "rsa_pub_dec" - rsa->meth->rsa_pub_dec is a NULL. pointer!
In contrary, RSA method rsa_priv_dec  points to to "pkcs11.dll ! pkcs11_rsa_priv_dec_method()".
![openssl_crash](https://cloud.githubusercontent.com/assets/13164219/19212251/b2cc7304-8d4d-11e6-9001-e3c76f2a17eb.gif)
Please correct this error, I really need working pkcs11 engine for my application, and I'm fed up with memory leaks in OpenSSL 1.0.2!

nased0:
OpenSSL crash when I execute command:
openssl s_client -msg -cert its1_crt.pem -key its1_key.pem -CAfile cck.pem -no_tls1_2 -connect 172.25.223.7:443
(this IP is behind VPN).
My openssl.cnf is:
[openssl.txt](https://github.com/OpenSC/libp11/files/517655/openssl.txt)

dengert:
In libp11/src/p11_rsa.c should line 432:
`ops = RSA_meth_new("libp11 RSA method", 0);`
be something like:
`ops = RSA_meth_dup(RSA_get_default_method());
RSA_meth_set1_name(ops, "libp11 RSA method");
RSA_meth_set_flags(ops, 0);`

(Have not tried this.)

nased0:
Thanks!
For now I can tell that with this patch engine pkcs11.dll works with the first command I mentioned (with PEM certs), I will test the second command at work, when I have access to the card reader.

mouse07410:
After applying the proposed patch on Mac OS X:
```
/Applications/Xcode.app/Contents/Developer/usr/bin/make  check-TESTS
FAIL: rsa-testpkcs11.softhsm
FAIL: rsa-testfork.softhsm
PASS: rsa-testlistkeys.softhsm
FAIL: rsa-evp-sign.softhsm
PASS: ec-testfork.softhsm
============================================================================
Testsuite summary for libp11 0.4.3_git
============================================================================
# TOTAL: 5
# PASS:  2
# SKIP:  0
[...]

Issue #123, opened 2016-10-13
Current master hangs up on PIN prompt (?)


mouse07410:
Mac OS X. Weird situation. I suspect it is related to cleaning the appropriate PKCS11 module. Would like help proving it (and obviously, fixing it :).

**Update**: it hangs *after* the requested operation (signature in this case) has been completed, so the signature file is produced and written to disk (before the application freezes up).

It hangs up when invoked "normally":
```
$ openssl dgst -engine pkcs11 -keyform engine -sign "pkcs11:manufacturer=piv_II;object=SIGN%20key;type=private" -sha384 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -out /tmp/derive.26192.text.sig /tmp/derive.26192.text
engine "pkcs11" set.
PKCS#11 token PIN: 
^C      [have to kill this hanging process]
$ 
```

When I invoke it with PKCSSPY, however, it completes correctly (`PKCS11SPY=/Library/OpenSC/lib/opensc-pkcs11.dylib`):
```
PKCS11_MODULE_PATH=/Library/OpenSC/lib/pkcs11-spy.dylib openssl dgst -engine pkcs11 -keyform engine -sign "pkcs11:manufacturer=piv_II;object=SIGN%20key;type=private" -sha384 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -out /tmp/derive.74923.text.sig /tmp/derive.74923.text


*************** OpenSC PKCS#11 spy *****************
Loaded: "/Library/OpenSC/lib/opensc-pkcs11.dylib"

0: C_GetFunctionList
2016-10-13 10:30:34.223
Returned:  0 CKR_OK

1: C_Initialize
2016-10-13 10:30:34.224
[in] pInitArgs = 0x7fff560da5b0
     flags: 2
       CKF_OS_LOCKING_OK
Returned:  0 CKR_OK

2: C_GetInfo
2016-10-13 10:30:34.953
[out] pInfo: 
      cryptokiVersion:         2.20
      manufacturerID:         'OpenSC Project                  '
      flags:                   0
      libraryDescription:     'OpenSC smartcard framework      '
      libraryVersion:          0.16
Returned:  0 CKR_OK
. . . . .
87: C_Sign
2016-10-13 10:30:37.731
[in] hSession = 0x7fdbb3004e00
[in] pData[ulDataLen] 00007fdbb160be30 / 256
[...]

PR #138, opened 2016-11-24
Corrected memory leak in pkcs11_CTX_free(PKCS11_CTX * ctx)

nased0:
I posted the screenshot of the leak at
https://github.com/OpenSC/libp11/issues/122#issuecomment-257850173
Now there are no leaks in pkcs11.dll (tested with Memory Validator and my app utilizing pkcs11 operations via cURL).

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -26,6 +26,9 @@ libp11_la_LDFLAGS = $(AM_LDFLAGS) \
 
 if HAVE_LD_VERSION_SCRIPT
 libp11_la_LDFLAGS += -Wl,--version-script="$(srcdir)/libp11.map"
+if WIN32
+libp11_la_LDFLAGS += -export-symbols "$(srcdir)/libp11.exports"
+endif
 else
 libp11_la_LDFLAGS += -export-symbols "$(srcdir)/libp11.exports"
 endif
--- a/src/p11_load.c
+++ b/src/p11_load.c
@@ -133,8 +133,7 @@ int pkcs11_CTX_reload(PKCS11_CTX * ctx)
  */
 void pkcs11_CTX_unload(PKCS11_CTX * ctx)
 {
-	PKCS11_CTX_private *cpriv;
-	cpriv = PRIVCTX(ctx);
+	PKCS11_CTX_private *cpriv = PRIVCTX(ctx);
 
 	/* Tell the PKCS11 library to shut down */
 	if (cpriv->forkid == _P11_get_forkid())
@@ -142,6 +141,7 @@ void pkcs11_CTX_unload(PKCS11_CTX * ctx)
 
 	/* Unload the module */
 	C_UnloadModule(cpriv->handle);
+	cpriv->handle = NULL;
 }
 
 /*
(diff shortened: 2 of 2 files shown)


mtrojnar:
Yes, we are aware of this (small) memory leak.  It prevents a deadlock on PKCS#11 modules that internally use the OpenSSL engine interface.
See https://github.com/OpenSC/libp11/issues/123 and https://github.com/OpenSC/libp11/commit/689bd3cc852bcd3e3e1c15f42949a4a577e29dbb for details.

nased0:
I tested my patch a few times with OpenSSL 1.0.2j (compiled by mingw)  and it does not hang on PIN prompt. I have successfully issued commands (on OpenSSL 1.0.2j): https://github.com/OpenSC/libp11/issues/122#issuecomment-262947160 as well as curl commands that ask for PIN:
curl --engine pkcs11 --tlsv1.1 --key-type ENG --cert-type ENG -k https://172.25.223.7:443/cepik/api/skp?wsdl
If my patch hangs the engine on a PIN prompt, please tell me the openssl command that hangs it.
I'm developing a service application, and any memory leaks are unacceptable for me.

mtrojnar:
Simple test cases may not catch this bug.  For example, with OpenSSL < 1.1.0 it requires locking callbacks to be initialized.  Single-threaded applications usually don't initialize locking callbacks.
Also, as I wrote, this bug only emerges with PKCS#11 modules that internally use the OpenSSL engine interface (for example SoftHSM2).

nased0:
Ok. I tested with 
openssl dgst -engine pkcs11 -keyform engine -sign "pkcs11:type=private" -sha384 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -out c:\tmp\derive.26192.text.sig

engine "pkcs11" set.
PKCS#11 token PIN:

and it hangs after asking for PIN.
But
openssl dgst -engine pkcs11 -keyform engine -sign "pkcs11:type=private;pin-value=1234" -sha384 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -out c:\tmp\derive.26192.text.sig

engine "pkcs11" set.

It does not ask for PIN and it hangs too!

mtrojnar:
Indeed.  The deadlock has nothing to do with asking for a PIN.

nased0:
I'm not sure that my command is correct, because on unmodified engine it hangs too!
P.S.
I use Encard Enigma card with SSL cert and pkey and USB reader, not SoftHSM2.
I will try to debug it at weekend.

nased0:
I tested this command on openssl 1.0.2j and openssl 1.1.0c:
openssl dgst -engine pkcs11 -keyform engine -sign "pkcs11:type=private;pin-value=1234" -sha384 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -out c:\tmp\derive.26192.text.sig
again and I realized that it does not hang, but waits for input!
Could you tell me what I'm doing wrong?
P.S.
Command OpenSSL req -engine pkcs11 -new -key "pkcs11:type=private;pin-value=1234"  -keyform engine -out req.pem -text -x509 -subj "/CN=Adam Smith" generates req.pem correctly, so i have no reasons to suspect that -key "pkcs11:type=private;pin-value=1234" is wrong.

mtrojnar:
What does it have to do with the memory leak?

nased0:
I want to verify existence of this deadlock, that prevents us from correcting the memory leak.

mtrojnar:
The proper thread to discuss this issue is #123, and the proper person @mouse07410, who reported it.

PR #139, opened 2016-11-30
Corrected memory leak in pkcs11_CTX_free v.2

nased0:
It works according to https://github.com/OpenSC/libp11/issues/123#issuecomment-264018817
P.S.
Removed "abominable" and unnecessary conversion :)

### DIFF
--- a/src/p11_load.c
+++ b/src/p11_load.c
@@ -159,8 +159,8 @@ void pkcs11_CTX_free(PKCS11_CTX * ctx)
 	if (cpriv->init_args) {
 		OPENSSL_free(cpriv->init_args);
 	}
-	if (cpriv->handle) {
-		pkcs11_CTX_unload(ctx);
+	if(cpriv->handle) {
+		OPENSSL_free(cpriv->handle);
 	}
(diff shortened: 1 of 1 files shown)


mtrojnar:
Does this PR really fix the memory leak?  It looks fairly harmless...

mtrojnar:
Please squash it into a single commit http://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit/

nased0:
It only fixes the memory leak (reported by Memory Validator) in data structure sc_pkcs11_module_t that contains the module handle.
It does not free the module. But this is a shared library, so repeated LoadLibrary or dlopen calls on that module should not increase memory usage. 
I would prefer proper finalization and unloading of a module, but it needs to be fixed in OpenSSL itself.
So it is the second best.

nased0:
About rebasing, I'm using only github's www interface, not a git console. And I'm currently at work.

mtrojnar:
Sure.  Take your time.  This is nothing urgent.

PR #145, opened 2017-02-06
Support custom user types

matthauck:
This somewhat-ugly patch is to enable support for custom user types,
used in particular by SafeNet HSM's proprietary CKU_CRYPTO_USER
(cf. https://github.com/gemalto/pycryptoki/blob/25bb878/pycryptoki/defines.py#L1516)

Note: I totally understand if you have no interest in this patch since it is supporting a proprietary flag that is not part of the standard and adds yet another dreaded engine ctrl command. I thought I might submit it as a PR after all just in case you or someone else might find it useful.

### DIFF
--- a/src/eng_back.c
+++ b/src/eng_back.c
@@ -231,6 +231,8 @@ int ctx_destroy(ENGINE_CTX *ctx)
 			CRYPTO_destroy_dynlockid(ctx->rwlock);
 #endif
 		OPENSSL_free(ctx);
+		// reset global user type flag
+		PKCS11_set_custom_user_type( 1UL ); // CKU_USER
 	}
 	return 1;
 }
@@ -902,6 +904,12 @@ static int ctx_ctrl_set_init_args(ENGINE_CTX *ctx, const char *init_args_orig)
 	return 1;
 }
 
+static int ctrl_set_user_type(ENGINE_CTX *ctx, long type)
+{
+	PKCS11_set_custom_user_type( (unsigned long) type );
+	return 1;
+}
+
 int ctx_engine_ctrl(ENGINE_CTX *ctx, int cmd, long i, void *p, void (*f)())
 {
 	(void)i; /* We don't currently take integer parameters */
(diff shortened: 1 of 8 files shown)


mtrojnar:
```C
int pkcs11_login(PKCS11_SLOT *slot, int so, const char *pin, int relogin)
```
The `so` parameter is defined as accepting either zero or nonzero, but I'm pretty sure all applications use either 0 or 1.  What if we defined other acceptable values?  Something like:
```C
switch (so) {
    case 0:
        user_type = CKU_USER;
        break;
    case 1:
        user_type = CKU_SO;
        break;
    default:
        user_type = so;
}
```
, or maybe more conservatively:
```C
switch (so) {
    case 0:
        user_type = CKU_USER;
        break;
    case 0x80000001: /* proprietary CKU_CRYPTO_USER */
        user_type = 0x80000001;
        break;
    default:
        user_type = CKU_SO;
}
```
What do you think?

matthauck:
That could work, though I'd be a bit concerned about:

* whether users of the api had used some other non-1 value assuming it was boolean
* the fact that `int` is probably not big enough to store an `unsigned long` value. 
* how one would go about setting this value via the engine interface.

matthauck:
Thought of another reason why that implementation would be confusing: the actual value of `CKU_SO` is 0 and `CKU_USER` is 1... 😕

mtrojnar:
> whether users of the api had used some other non-1 value assuming it was boolean

True.  This is why I proposed the conservative solution.

> the fact that `int` is probably not big enough to store an `unsigned long` value.

Surprisingly, `int` and `unsigned long` are the same size on many platforms, including 64-bit Windows.  On all platforms supported by OpenSSL `int` is big enough to hold `0x80000001` (often as a negative value).  It would be helpful to change "int" to "unsigned int" to avoid unsigned/signed casting while preserving binary compatibility.

> how one would go about setting this value via the engine interface.

Implementing an engine ctrl is indeed a separate issue.

> Thought of another reason why that implementation would be confusing: the actual value of `CKU_SO` is 0 and `CKU_USER` is 1...

Yes, this is precisely the reason for `switch (so)`.

Okay, you convinced me there is no clean way to implement this hack.  I'm closing this PR.

matthauck:
ok. so be it. :)

PR #149, opened 2017-03-03
Fix #148: Remove slot-specific lock

### LINKED ISSUE #148: Locking issue when loading key and signing at the same time
We have a test that spins up lots of threads and on each of those threads: (1) loads a key with libp11, and (2) signs something with that key (via engine api). I recently incorporated the patch for #144 into the libp11 we're using and discovered a locking problem that 9bbdd80 may have introduced. 

I think this was not a problem before since keys/certs were only loaded once, and with that change they are loaded often. 

The basic problem is that there are multiple locks in play here. The context rwlock is used in [p11_key.c](https://github.com/OpenSC/libp11/blob/9bbdd80/src/p11_cert.c#L50) and [p11_cert.c](https://github.com/OpenSC/libp11/blob/9bbdd80/src/p11_cert.c#L50), however the slot rwlock is used in [p11_rsa.c](https://github.com/OpenSC/libp11/blob/9bbdd80/src/p11_rsa.c#L98), [p11_ec.c](https://github.com/OpenSC/libp11/blob/9bbdd80/src/p11_ec.c#L307), etc. 

Since these calls are not being synchronized with each other, it is not difficult to get into the state where the "load keys" or "load certs" operation comes back with CKR_OPERATION_ACTIVE (testing with softhsm). 

It seems to me that only one lock should be used throughout libp11, and suggest it should probably be the context lock. Trying to lock both locks could easily get you into a deadlock scenario, and keeping both locks around will likely bring back issues like this in the future.  

Thoughts?

matthauck:
Having two locks leaves operations in the pkcs11 module improperly
synchronized, e.g.when loading keys and signing at the same time.

### DIFF
--- a/src/libp11-int.h
+++ b/src/libp11-int.h
@@ -65,9 +65,6 @@ typedef struct pkcs11_slot_private {
 	/* options used in last PKCS11_login */
 	char *prev_pin;
 	int prev_so;
-
-	/* per-slot lock */
-	PKCS11_RWLOCK rwlock;
 } PKCS11_SLOT_private;
 #define PRIVSLOT(slot)		((PKCS11_SLOT_private *) ((slot)->_private))
 #define SLOT2CTX(slot)		(PRIVSLOT(slot)->parent)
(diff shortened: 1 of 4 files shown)


mtrojnar:
Good idea.  The potential performance gain due to more granular locks was questionable anyway...

Issue #146, opened 2017-02-09
Expose a way to set the UI_METHOD for a engine-loaded private key


matthauck:
See discussion: https://github.com/OpenSC/libp11/commit/be22da7#commitcomment-20772428

mtrojnar:
Implemented: https://github.com/OpenSC/libp11/commit/ef756be48e7b920eb2d809ba23a6bb06c1571bf9

matthauck:
This can probably be closed now, right?

mtrojnar:
You requested this feature, so I presume you can confirm that it works as you expected.  I would appreciate your test results.

matthauck:
Sorry, been tied up with other things. Getting around to testing this shortly.

matthauck:
Hmm. Sadly there does not appear to be enough information in the UI prompt to decide which PIN to use: [p11_key.c#L365-L369](https://github.com/OpenSC/libp11/blob/ef756be48e7b920eb2d809ba23a6bb06c1571bf9/src/p11_key.c#L365-L369)

We need to be able to support keys from different slots, which may each have different PINs. Ideally, the prompt would print out the key URI or something so that the consumer would know which PIN to provide.

mtrojnar:
Doesn't the user already know the selected key from the calling application?

Otherwise, it would be nice to have some form of key identifier in the UI prompt.

matthauck:
well, the engine interface only allows specifying one UI method, whereas the ENGINE_load_private_key openssl calls allows for different pins for different keys (i.e slots).

mtrojnar:
Does this issue affect your deployment?

matthauck:
Yes. We have multiple keys that we load from different slots.

mtrojnar:
What does your application do?  Couldn't it display a message to the user before a private key is used?

matthauck:
I think there's a bit of misunderstanding here. When libp11 needs the `CKU_CONTEXT_SPECIFIC` pin for a particular key (when there are, say, 2 keys in use), it calls out to the singleton UI_METHOD I've specified via the engine api, how do I as the application know which key/slot this pin is for? The application can know both pins, and which pin is for which key, but the UI* is not passing enough information for the application to know which pin is being requested. 

Does that make sense?

mtrojnar:
Yes, it does.  I'm trying to understand the specific application that randomly makes signatures with two separate private keys on two separate tokens that both use `CKU_CONTEXT_SPECIFIC` PINs.

[...]

PR #152, opened 2017-04-06
Add more information to UI prompt

matthauck:
This alerts the user as to which token/key a PIN is being requested for.
This addresses some remaining concerns discussed in issue #146.

### DIFF
--- a/src/eng_back.c
+++ b/src/eng_back.c
@@ -114,22 +114,7 @@ static int ctx_get_pin(ENGINE_CTX *ctx, const char* token_label, UI_METHOD *ui_m
 	memset(ctx->pin, 0, MAX_PIN_LENGTH+1);
 	ctx->pin_length = MAX_PIN_LENGTH;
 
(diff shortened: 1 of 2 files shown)


matthauck:
This changes prompt texts like:

> PKCS#11 token PIN:
to:
> Enter PIN for PKCS#11 token 'my-slot':

and
> PKCS#11 key PIN:
to
> Enter PIN for PKCS#11 key 'my-key', token 'my-slot':

While also running the strings through `UI_construct_prompt`, so that the ui-method can reconstruct it differently if they so desire.

matthauck:
Yeah, good points. I felt kinda dirty doing all that string concatenation. This is much cleaner. 

Please take another look.  Should now look like:

> Enter PKCS#11 token PIN for my-slot:
> Enter PKCS#11 key PIN for my-key:

mtrojnar:
Yes, this is exactly what I meant.  Thank  you.

PR #161, opened 2017-06-22
Use pkcs11_find_key to select key with the same id as the first certi…

nased0:
…ficate (if no key id is given).

This is my solution to a real world problem - I got a card with two keys and one certificate in slot 0, token 0 and the engine was unable to select the right key - the one with the same id as the certificate. Fortunately I found the function pkcs11_find_key that does exactly that.

My program is supposed to work when the user inserts his card into the reader, without listing the contents of the card by using the pkcs11-tool and copying and pasting the correct key id from Windows console into a configurations window. Try to explain that, step by step by phone, to the car diagnostician!

### DIFF
--- a/src/eng_back.c
+++ b/src/eng_back.c
@@ -27,6 +27,7 @@
  */
 
 #include "engine.h"
+#include "libp11-int.h"
 #include <stdio.h>
 #include <string.h>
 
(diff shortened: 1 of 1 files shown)


mtrojnar:
The proper solution is to select the PKCS#11 object by its user-friendly label rather than by its obscure id.

I reject this PR because it is confusing and only useful for your very specific use case (only one certificate and multiple private keys).  Applications should select specific objects rather than depend on the device to only contain a single valid object.

The PR also breaks separation between the OpenSSL engine and libp11 by invoking the internal `pkcs11_find_key()` function.  This, however, could easily be fixed with its exported `PKCS11_find_key()` version.

PR #168, opened 2017-07-18
Move manual configuration up to the configuration file section and

bmomjian:
update the arguments to match the configuration file example and current
usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ to copy the engine_pkcs11 to that location as "libpkcs11.so" to ease usage.
 This is handle by 'make install' of engine_pkcs11.
 
 
-## Using the engine from the command line tool
+## Using the engine from the command line
 
 In systems with p11-kit-proxy engine_pkcs11 has access to all the configured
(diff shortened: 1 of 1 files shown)


bmomjian:
This URL:

   https://github.com/OpenSC/libp11

explains how to configure the OpenSSL config file to know about the libraries under the heading "Using in systems without p11-kit,", but at the bottom it explains that you have to "provide the engine configuration explicitly." I think the second part is unnecessary if the first part is done properly.

mtrojnar:
I agree with adding a clear note that you **either** need to add a configuration file, or commandline options.  Even though this part of the documentation also applies to configuring the engine, I think it fits better in the section on using OpenSSL commandline tools.

bmomjian:
Are you saying this should be moved into the section on the same page titled, "Using the engine from the command line tool", or to some other page?  If so, what other page?

mtrojnar:
same page

bmomjian:
OK, I changed some headings to put it all under "command-line," and adjusted some wording.

PR #172, opened 2017-07-19
Choose OpenSSL version for correct library name

beamer159:
OpenSSL 1.1.0 changed its library file name from libeay32 to libcrypto. This PR allows the user to specify they are using OpenSSL 1.1.0+ with OPENSSL_VERSION. Setting this will tell the makefile to look for libcrypto instead of libeay32. If the user does not set OPENSSL_VERSION, the library will build as before.


mtrojnar:
DLL name autodetection was implemented in https://github.com/OpenSC/libp11/commit/e5559b35a6e76b3e1e09214ea3df9dee29321ea3.

PR #182, opened 2017-10-11
Fix OpenSSL version check for EVP_PKEY_meth_get_XXX()

mouse07410:
These accessors have been added to OpenSSL-1.0.2m+. Thus defining `static void EVP__KEY_meth_get_sign()` breaks the compilation on OpenSSL-1.0.2m-dev. This PR fixes that problem.


mouse07410:
@mtrojnar ?

mouse07410:
Thanks!

Issue #184, opened 2017-10-13
Error storing private key on bit4id smartcard


n3wtron:
I've got a "Vendor General Error" storing a private key on a smartcard (vendor:bit4id) calling PKCS11_store_private_key.

The pkcs11-tool (from terminal) works so I've tried to compare the libp11 code to pkcs11-tool.
With this modification https://github.com/n3wtron/libp11/tree/feature/import_priv_key_exponents it works, but I'm not so expert to say that is the correct solution.

I'll really appreciate if someone has some time to check it and give me some feedback

Thank you

dengert:
A minor comment is the choice of variables names. For readability, I would suggest
rsa_exp1, rsa_exp2 and rsa_coef be renamed  rsa_dmp1, rsa_dmq1 and rsa_iqmp. 

(I have no way to test this.)

dengert:
Based on looking at the OpenSSL code, I would also replace `if (rsa_exp1 && rsa_exp2 && rsa_coef)` with 3 separate tests and only add the attributes for the ones present.

mtrojnar:
@n3wtron Makes sense.  Please apply the changes suggested by @dengert and submit a PR.

n3wtron:
@mtrojnar, I will modify the code and submit a merge request. Thank you

mtrojnar:
#186 has been merged.

PR #186, opened 2017-10-17
Fix store private key on bit4id card

n3wtron:
Relative to #184


n3wtron:
I don't think that the CI error on x86 windows platform is related to my modifications

PR #187, opened 2017-10-17
Add remove key/certificate functionality

n3wtron:


n3wtron:
I'm developing a GUI application (with QT) to manage smartcards.
Some features:
- import p12
- manage certificates and keys (including deletion)

PR #194, opened 2017-11-10
change documentation of PKCS11_change_pin

kiranpradeep:
PKCS11_change_pin can change both USER and SO pin. Updated documentation to reflect the same.

What this pull request lacks is, it doesn't tell that, pre-login is mandatory for the PKCS11_change_pin call. I am not sure how to reflect that in doc.


mtrojnar:
What about "Change the currently used (either USER or SO) PIN on a token."?

PR #200, opened 2018-01-09
rsa: fix write to invalid pointer

emmanuel-deloget:
BIGNUM * shall be either a valid BIGNUM instance or NULL before calling
key_getattr_bn() otherwise the software will crash.

This bug can be reproduced with PKCS#11 providers who does not provide
the public exponent on private keys.

Signed-off-by: edt <emmanuel.deloget@eho.link>


emmanuel-deloget:
Sorry, identify issue (I'm not myself?). 

I respin this in a few hours.

mtrojnar:
That's a good point.  BN_bin2bn() indeed requires either a valid BIGNUM or NULL.

PR #207, opened 2018-02-03
Fixed Bufferoverflow in Listkeys Example

wirsiing:
As said in Issue199, there was a buffer overflow in the example file listkeys with password[20].
I now fixed this issue (very minor change) per your request.

### DIFF
--- a/examples/listkeys.c
+++ b/examples/listkeys.c
@@ -40,7 +40,7 @@ int main(int argc, char *argv[])
 	PKCS11_SLOT *slots=NULL, *slot;
 	PKCS11_KEY *keys;
 	unsigned int nslots, nkeys;
-	char password[20];
+	char *password;
 	int rc = 0;
 
 	if (argc < 2) {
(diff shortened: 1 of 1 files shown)


examples/listkeys.c
```diff
@@ -83,6 +83,7 @@ int main(int argc, char *argv[])
 	list_keys("Public keys", keys, nkeys);
 
 	if (slot->token->loginRequired && argc > 2) {
+		password = malloc(sizeof(argv[2])+1);
```
FdaSilvaYY: s/sizeof/strlen 
... else the fix will be worst than the initial version...
Where/when will it be free ?

wirsiing: You are correct. Sorry for the mistake.

I guess one could free after memset in line 92?

mtrojnar:
I took the liberty of fixing it with https://github.com/OpenSC/libp11/commit/4806f1efb0eb8d7150749948e62ff0c11a88b5db.
Also see: https://en.wikipedia.org/wiki/KISS_principle.

Issue #209, opened 2018-03-18
engine_pkcs11 segfault in EC operations with SafeNet


aalba6675:
Using SafeNet as OpenSSL engine via engine_pkcs11: EC operations are segfaulting with EC keys when no public key is on token. Using the gem engine from SafeNet works correctly.

It would appear that an EC public key with the same ID is required.

RSA operations work when there is no corresponding public key on token.

```
# rpm -q openssl engine_pkcs11
openssl-1.0.2k-8.el7.x86_64
engine_pkcs11-0.4.7-1.el7.x86_64
```

Failed key
```
openssl pkey -pubout -engine pkcs11 -inform engine -in 'pkcs11:token=WEBTOKEN;object=webkey_1'
luna_get_conf_path: "/etc/Chrystoki.conf"
LibPath64: "/usr/lib/libCryptoki2_64.so"
EnableEcdsaEx: "1"
LogLevel: "7"
DisableDsa: "0"
DisableEcdsa: "0"
DisableRand: "0"
DisableSessionCache: "1"
DisablePublicCrypto: "1"
EnableLoadPrivKey: "1"
EnableLoadPubKey: "1"
EnableLoginInit: "1"
EnableRsaGenKeyPair: "1"
EnableDsaGenKeyPair: "1"
EnableRsaSignVerify: "1"
engine "pkcs11" set.
Missing CKA_ALWAYS_AUTHENTICATE attribute
Segmentation fault
```

aalba6675:
p11tool shows the difference in keys:
```
Object 1:
        URL: pkcs11:token=WEBTOKEN;object=webkey_1;type=private
        Type: Private key
        Label: webkey_1
        Flags: CKA_WRAP/UNWRAP; CKA_PRIVATE; CKA_SENSITIVE; 
        ID: 00:00:01:5e:88:d2:7e:b8

Object 2:
        URL: pkcs11:token=WEBTOKEN;object=webkey_2;type=private
        Type: Private key
        Label: webkey_2
        Flags: CKA_PRIVATE; CKA_SENSITIVE; 
        ID: 50:00:00
```

aalba6675:
bt:

```
Program received signal SIGSEGV, Segmentation fault.
EC_POINT_point2oct (group=0x7acfc0, point=0x0, form=POINT_CONVERSION_UNCOMPRESSED, buf=buf@entry=0x0, len=len@entry=0, ctx=ctx@entry=0x0) at ec_oct.c:144
[...]

PR #210, opened 2018-03-20
src/p11_ec.c: make additional check for EC_POINT

aalba6675:
[UPDATE] the AppVeyor failure doesn't seem to be related to the source code change

-------------

TL;DR Add a non-standard check for EC_POINT in a EC private key to support vendor-optimization where private key has EC_PARAMS and EC_POINT attributes.

1. For EC private keys we make a round-trip to the HSM to get EC_PARAMS and EC_POINT by looking for a corresponding public key object.
1. Some non-standard but helpful modules include EC_PARAMS/EC_POINT with the private key (SafeNet)
1. We already have a check for EC_PARAMS in the code (irregardless of public or private key added by  @mtrojnar), so I hope it is not too intrusive to make one more check for EC_POINT to support this nice vendor-optimization (saving a round-trip to the HSM)
1. This optimization is useful as it saves having an extra (relatively useless, duplicates information in a cert, for example) object on token ($$$)

Note: [Vendor Quirk] This is what AWS CloudHSM engine is doing: without a corresponding EC public key object OpenSSL apps like `req` and `pkey -pubout` work as the engine has filled in the public key attribute using the quirk of the private key

With engine_pkcs11, both these operations segfault (as expected) as the public key attribute is NULL.

The investigation is at #209.

### DIFF
--- a/src/p11_ec.c
+++ b/src/p11_ec.c
@@ -208,28 +208,39 @@ static EC_KEY *pkcs11_get_ec(PKCS11_KEY *key)
 		OPENSSL_free(params);
 	}
 
-	/* Now retrieve the point */
-	pubkey = key->isPrivate ? pkcs11_find_key_from_key(key) : key;
-	if (pubkey == NULL)
-		return ec;
-	if (!key_getattr_alloc(pubkey, CKA_EC_POINT, &point, &point_len)) {
-		const unsigned char *a;
-		ASN1_OCTET_STRING *os;
+	/* Now retrieve the point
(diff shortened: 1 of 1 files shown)


mtrojnar:
Can you please check if https://github.com/OpenSC/libp11/commit/281ccb3d60e77366bd6ece4b8a18d5ef28cb4f45 does the job for you?

aalba6675:
I can confirm that this commit works.

mtrojnar:
Thank you.

PR #216, opened 2018-04-04
travis: fix ci build

williamcroberts:
OpenSSL doesn't have EDDSA support, so disable it when building SoftHSMv2 via
the --disable-eddsa option.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -45,6 +45,6 @@ mkdir prerequisites
 cd prerequisites
 install_from_github OpenSC OpenSC master
 # softhsm is required for "make check"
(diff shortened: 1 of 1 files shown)


williamcroberts:
OpenSC has a build issue, I submitted a PR to them:
https://github.com/OpenSC/OpenSC/pull/1311

This gets travis running again, and we can see clang passing.

mtrojnar:
Current master builds fine without this fix: https://travis-ci.org/OpenSC/libp11/builds/362267951

SoftHSM2 doesn't currently have the --disable-eddsa option.  It looks like a forced push on https://github.com/opendnssec/SoftHSMv2/commits/develop.  I switched to using the master branch instead.

Issue #217, opened 2018-04-05
Current master fails fork test with OpenSSL-1.1.1


mouse07410:
MacOS 10.13.4, Xcode-9.3, current master OpenSSL (`OpenSSL 1.1.1-pre5-dev  xx XXX xxxx`). Problem seems to be with generating random numbers on a token:

```
$ ./configure --disable-silent-rules --prefix=/Users/uri/openssl-1.1 --with-pkcs11-module=/opt/local/lib/p11-kit-proxy.dylib --with-enginesdir=/Users/uri/openssl-1.1/lib/engines-1.1/
. . . . .
/Applications/Xcode.app/Contents/Developer/usr/bin/make  check-TESTS
PASS: rsa-testpkcs11.softhsm
PASS: rsa-testfork.softhsm
PASS: rsa-testlistkeys.softhsm
PASS: rsa-evp-sign.softhsm
PASS: ec-testfork.softhsm
FAIL: fork-change-slot.softhsm
============================================================================
Testsuite summary for libp11 0.4.8_git
============================================================================
# TOTAL: 6
# PASS:  5
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
============================================================================
See tests/test-suite.log
============================================================================
make[3]: *** [test-suite.log] Error 1
make[2]: *** [check-TESTS] Error 2
make[1]: *** [check-am] Error 2
make: *** [check-recursive] Error 1
$ cat tests/test-suite.log 
============================================
   libp11 0.4.8_git: tests/test-suite.log
============================================

# TOTAL: 6
# PASS:  5
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: fork-change-slot.softhsm
==============================

Current directory: /Users/uri/src/libp11/tests
Source directory: .
Output directory: output.76528
-n * Initializing smart card... 
ok
-n * Initializing smart card... 
ok
Key pair generated:
Private Key Object; RSA 
  label:      pkey
  ID:         01020304
  Usage:      decrypt, sign, unwrap
Public Key Object; RSA 1024 bits
  label:      pkey
  ID:         01020304
  Usage:      encrypt, verify, wrap
Found token (39dc8410-4c78-0dbe-dd48-bbf84b72866b) with matching token label.
[...]

PR #222, opened 2018-04-09
random support: make configurable

williamcroberts:
Make random support configurable and make the default OFF.

Modules that use OpenSSL will conflict on random number
generation, and either one of 2 things will happen:
1. Random generation will end up in an infinite loop
   between the PKCS#11 module and the pkcs#11 engine.
2. The PKCS#11 module will change the random engine
   thus preventing issue 1, but causing the random
   source to be changed.

Fixes: #217

Signed-off-by: William Roberts <william.c.roberts@intel.com>

### DIFF
--- a/src/eng_back.c
+++ b/src/eng_back.c
@@ -52,6 +52,7 @@ struct st_engine_ctx {
 	UI_METHOD *ui_method;
 	void *callback_data;
 	int force_login;
+	int enable_rand;
 
 	/* Engine initialization mutex */
 #if OPENSSL_VERSION_NUMBER >= 0x10100004L && !defined(LIBRESSL_VERSION_NUMBER)
@@ -963,6 +964,12 @@ static int ctx_ctrl_force_login(ENGINE_CTX *ctx)
 	return 1;
 }
 
+static int ctx_ctrl_enable_rand(ENGINE_CTX *ctx) {
+
+	ctx->enable_rand = 1;
+	return 1;
+}
+
 int ctx_engine_ctrl(ENGINE_CTX *ctx, int cmd, long i, void *p, void (*f)())
 {
 	(void)i; /* We don't currently take integer parameters */
@@ -989,6 +996,8 @@ int ctx_engine_ctrl(ENGINE_CTX *ctx, int cmd, long i, void *p, void (*f)())
 		return ctx_ctrl_set_callback_data(ctx, p);
 	case CMD_FORCE_LOGIN:
 		return ctx_ctrl_force_login(ctx);
+	case CMD_ENABLE_RAND:
+		return ctx_ctrl_enable_rand(ctx);
 	default:
 		ENGerr(ENG_F_CTX_ENGINE_CTRL, ENG_R_UNKNOWN_COMMAND);
 		break;
@@ -1040,5 +1049,11 @@ int rand_bytes (unsigned char *buf, int num)
 	/* Goes back to openssl, where 1 is success */
 	return 1;
 }
+
+int ctx_is_rand_enabled(ENGINE_CTX *ctx)
+{
+	return ctx->enable_rand;
+}
+
 
 /* vim: set noexpandtab: */
--- a/src/eng_back.c
+++ b/src/eng_back.c
@@ -241,7 +241,15 @@ ENGINE_CTX *ctx_new()
 	ctx->rwlock = CRYPTO_get_dynlock_create_callback() ?
 		CRYPTO_get_new_dynlockid() : 0;
 #endif
-
+/*
+ * OpenSSL Version prior to 1.1.0 are known to work
+ * with random.
+ */
+#if OPENSSL_VERSION_NUMBER < 0x101010000
+	ctx->enable_rand = 1;
+#else
+	ctx->enable_rand = 0;
+#endif
 	return ctx;
 }
 
@@ -964,10 +972,13 @@ static int ctx_ctrl_force_login(ENGINE_CTX *ctx)
 	return 1;
 }
 
-static int ctx_ctrl_enable_rand(ENGINE_CTX *ctx) {
+static int ctx_ctrl_enable_rand(ENGINE_CTX *ctx, long value) {
 
-	ctx->enable_rand = 1;
-	return 1;
+	if (value == 0 || value == 1) {
+		ctx->enable_rand = (int)value;
+		return 1;
+	}
+	return 0;
 }
 
 int ctx_engine_ctrl(ENGINE_CTX *ctx, int cmd, long i, void *p, void (*f)())
@@ -997,7 +1008,7 @@ int ctx_engine_ctrl(ENGINE_CTX *ctx, int cmd, long i, void *p, void (*f)())
 	case CMD_FORCE_LOGIN:
 		return ctx_ctrl_force_login(ctx);
 	case CMD_ENABLE_RAND:
-		return ctx_ctrl_enable_rand(ctx);
+		return ctx_ctrl_enable_rand(ctx, i);
 	default:
 		ENGerr(ENG_F_CTX_ENGINE_CTRL, ENG_R_UNKNOWN_COMMAND);
 		break;
--- a/README.md
+++ b/README.md
@@ -166,6 +166,7 @@ The supported engine controls are the following.
 * **SET_USER_INTERFACE**: Set the global user interface
 * **SET_CALLBACK_DATA**: Set the global user interface extra data
 * **FORCE_LOGIN**: Force login to the PKCS#11 module
+* **ENABLE_RAND**: Enable support of random number generation from the PKCS#11 module
 
 An example code snippet setting specific module is shown below.
 
--- a/README.md
+++ b/README.md
@@ -166,7 +166,12 @@ The supported engine controls are the following.
 * **SET_USER_INTERFACE**: Set the global user interface
 * **SET_CALLBACK_DATA**: Set the global user interface extra data
 * **FORCE_LOGIN**: Force login to the PKCS#11 module
-* **ENABLE_RAND**: Enable support of random number generation from the PKCS#11 module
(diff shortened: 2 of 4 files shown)


mouse07410:
In general having it dynamically configurable (setting at run-time) is nice.

However, how would it work for, e.g., `openssl rand -engine pkcs11 -hex 32`?

williamcroberts:
@mouse07410 

This dynamic switch is logically no different then conditionally compiling it in for OSSL 1.0.2. If we want to, we could switch the default to ON for OSSL 1.0.2 and off for others. But from a user-perspective, this makes it harder as you don't know the state of the software given the software's version. Ie version X of the engine could have different behavior given the linked version of OSSL. If you have a default, people know what to expect unless they tinker with it. Whether or not it works for their particular setup, that's a separate issue. We could just mark this in the readme as experimental and what it is known to work on.

You'll need a way to run time disable it anyways when your loaded PKCS#11 module is calling into OSSL and has engine conflicts.

I don't really understand what you mean by "how would it work for 'openssl rand -engine pkcs11 -hex 32`?
If you mean, how do you set this value, you do it in an openssl conf engine like so:
`OPENSSL_CONF=~/engine.conf openssl rand -engine pkcs11 32 -hex`

mouse07410:
> This dynamic switch is logically no different then conditionally compiling it in for OSSL 1.0.2.

It certainly is different - because conditionally compiled code just works (assuming the defines were right) so the user doesn't need to know about it, and with the dynamic switch the user has to set it.

> If we want to, we could switch the default to ON for OSSL 1.0.2 and off for others.

Until there's a fix for OSSL-1.1.x yes that's exactly what we want. And after it is fixed - I see no reason to disable it ever.

> But from a user-perspective, this makes it harder as you don't know the state of the software given the software's version.

Normally, you have to recompile libp11 as a new release of OSSL comes out, to be safe.

> You'll need a way to run time disable it anyways when your loaded PKCS#11 module is calling into OSSL and has engine conflicts.

I thought the whole point of this exercise was to get rid of the engine conflicts?

> I don't really understand what you mean by "how would it work for `openssl rand -engine pkcs11 -hex 32`"?

I meant how exactly would the user turn this switch on or off, especially assuming that the user wants to use pkcs11 engine to access RSA and/or ECC keys on his PKCS#11 token, and the only question is whether to delegate `RAND_bytes()` to this token as well.

> If you mean, how do you set this value, you do it in an openssl conf engine like so: `OPENSSL_CONF=~/engine.conf openssl rand -engine pkcs11 32 -hex`

And what would be in `engine.conf` file? What parameter would tell the engine that it should hook RSA operations, but not RAND, or both RSA and RAND? And how would the engine get that parameter? I didn't think the engine read config files...

williamcroberts:
> -----Original Message-----
> From: Mouse [mailto:notifications@github.com]
> Sent: Tuesday, April 10, 2018 5:08 PM
> To: OpenSC/libp11 <libp11@noreply.github.com>
> Cc: Roberts, William C <william.c.roberts@intel.com>; Author
> <author@noreply.github.com>
> Subject: Re: [OpenSC/libp11] random support: make configurable (#222)
> 
> 	This dynamic switch is logically no different then conditionally compiling it
> in for OSSL 1.0.2.
> 
> It certainly is different - because conditionally compiled code just works

That's simply not true.

> (assuming the defines were right) so the user doesn't need to know about it, and
> with the dynamic switch the user has to set it.

Not if you set the default to a conditionally compiled flag. Things start to get
complicated when it CAN work on OSSL v1.2, but we never patched our
code to compile this back in. Users downstream would have to wait for
a new release.

Which is bad if the user can't bank on expected behavior. If we did this,
we would need to change the dynamic switch to take a long value, ie 0 or 1.
Which would alleviate the issue with the test code not liking non argument
based options in the config file. Not sure why that is.

> 
> 	If we want to, we could switch the default to ON for OSSL 1.0.2 and off
> for others.
> 
> Until there's a fix for OSSL-1.1.x yes that's exactly what we want. And after it is
> fixed - I see no reason to disable it ever.
> 
> 	But from a user-perspective, this makes it harder as you don't know the
> state of the software given the software's version.
> 
> Normally, you have to recompile libp11 as a new release of OSSL comes out, to be
> safe.
> 
> 	You'll need a way to run time disable it anyways when your loaded
> PKCS#11 module is calling into OSSL and has engine conflicts.
> 
> I thought the whole point of this exercise was to get rid of the engine conflicts?
> 
> 	I don't really understand what you mean by "how would it work for
> openssl rand -engine pkcs11 -hex 32"?
> 
> I meant how exactly would the user turn this switch on or off, especially assuming
> that the user wants to use pkcs11 engine to access RSA and/or ECC keys on his
> PKCS#11 token, and the only question is whether to delegate RAND_bytes() to
> this token as well.
> 
> 	If you mean, how do you set this value, you do it in an openssl conf
> engine like so: OPENSSL_CONF=~/engine.conf openssl rand -engine pkcs11 32 -
> hex
> 
> And what would be in engine.conf file? What parameter would tell the engine
> that it should hook RSA operations, but not RAND, or both RSA and RAND? And
> how would the engine get that parameter? I didn't think the engine read config
> files...

Engines can take parameters via OSSL config files. Engines can register for commands
and they get routed to an engine control function:
https://github.com/OpenSC/libp11/blob/master/src/eng_front.c#L214

The patch actually adds a new control parameter, which are documented
here: https://github.com/OpenSC/libp11/blob/master/README.md#engine-controls

There's an example config file here, that is used in the tests:
https://github.com/OpenSC/libp11/blob/master/tests/engines.cnf.in

src/eng_back.c
```diff
@@ -241,7 +241,13 @@ ENGINE_CTX *ctx_new()
 	ctx->rwlock = CRYPTO_get_dynlock_create_callback() ?
 		CRYPTO_get_new_dynlockid() : 0;
 #endif
-
+/*
+ * OpenSSL Version prior to 1.1.0 are known to work
+ * with random.
+ */
+#if OPENSSL_VERSION_NUMBER < 0x101010000
```
mouse07410: This looks reasonable

src/eng_back.c
```diff
@@ -241,7 +241,13 @@ ENGINE_CTX *ctx_new()
...
+	ctx->enable_rand = 1;
+#endif
```
mouse07410: Shouldn't `ctx->enable_rand` be explicitly set to zero otherwise?

mouse07410:
How would the engine invoked via `p11-kit` receive this parameter? Here's an example of how a token is used with p11-kit + OSSL command line:
```
openssl pkeyutl -engine pkcs11 -keyform engine -decrypt -inkey "pkcs11:manufacturer=piv_II;object=KEY%20MAN%20key;type=private" -pkeyopt rsa_padding_mode:oaep -in /tmp/derive.104.key.enc
```

Are you saying that with your patch the calling application would have to explicitly pass the config to the engine, or invoke an engine control command?

src/eng_back.c
williamcroberts: malloc + memset to zero, see: https://github.com/OpenSC/libp11/blob/master/src/eng_back.c#L221-L224

williamcroberts:
`OPENSSL_CONF=foo.cnf openssl pkeyutl -engine pkcs11 -keyform engine -decrypt -inkey "pkcs11:manufacturer=piv_II;object=KEY%20MAN%20key;type=private" -pkeyopt rsa_padding_mode:oaep -in /tmp/derive.104.key.enc`

src/eng_back.c
williamcroberts: I'm not sure if I have this version number right. I checked out 1.1.0 and used that version number...so I think I do.

I wonder if the issues we're seeing has to do with:
https://github.com/openssl/openssl/blob/master/crypto/rand/rand_lib.c#L817
Ihaven't looked at it in detail though.

src/eng_back.c
mtrojnar: Explicit is better than implicit.  Code clarity is more important than a tiny performance optimization

mtrojnar: Can't you see that setting `ctx->enable_rand = 0;` on `#else` would improve code readability?  @mouse07410 is right here.

williamcroberts: No, especially when the initialization routine is inconsistent in it's practice and about 20 lines before is a memset to zero.
Why memset the memory to write a zero to it? Why malloc + memset when _zalloc is available?

The block from 227-236 is incredibly convoluted to read. Conditional code is harder to read, especially runtime conditional code mixed with compile time conditional code. That should probably be moved to it's own routine.

With that said will I do it, sure.

ansasaki:
This pull request do not contain tests covering the implemented feature. Please, add tests to your pull request.

williamcroberts:
@ansasaki The tests for random already exist, this is working around a bug for those downstream. It's a bit limited what OSSL versions we test against due to travis.

I could add a test to disable rand via config and watch it fall flat on its face, would that be sufficient?

ansasaki:
@williamcroberts I don't see specific random number generation tests in the test suite. Some tests call the random number generation, but it would be interesting to have tests specifically for it.

> I could add a test to disable rand via config and watch it fall flat on its face, would that be sufficient?

It is a good practice to always add tests to cover the changes you are proposing. Since your PR proposes to make it configurable, it would be interesting to have tests which exercises this.

I don't see why you have to try to run and fail. You could enable the interface, load the engine and check if the random methods changed. And the other way around.

williamcroberts:
Yeah it's good practice, agreed there. I don't see any unit-level tests that I can add to. Unless I am missing them, so i'd have to create a  unit testing framework for this project, time which I don't have. 

fork-test.c calls into the EVP_Random() routine, which actually gets the random through softhsm. I would a need a test hat uses a config that disables it.

williamcroberts:
reported fixed: https://github.com/OpenSC/libp11/pull/224/files#diff-9b905ce79ee26163834b3e2a1832cb86L209

PR #223, opened 2018-04-16
Travis: Include Coverity Scan

frankmorgner:


mtrojnar:
This warning is an obvious false positive.  This input is as trusted as invoking `dlopen` directly.
This is what happens when you trust automated tools: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=363516

frankmorgner:
Yes, you can mock Coverity Scan and Debian. But you can also take this hint serious and look at something like [POSIX portable filename recommendations](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282) or whatever the [system defines to be valid](https://en.wikipedia.org/wiki/Filename)...

mtrojnar:
Do you suggest that libp11 shoud restrict directory names to English alphabet?
While it is indeed a good idea for applications to only use portable names for their files, this good idea should never be **enforced** on users.  Also, an interface library, such as libp11, would be **the worst** possible place for enforcing file name format restrictions.

frankmorgner:
OK, then feel free to dismiss the issue on the Coverity Scan pages.

mtrojnar:
Wouldn't it be better to add an annotation tag, so it will benefit other Coverity users?
https://scan4.coverity.com/doc/en/cov_checker_ref.html#static_checker_TAINTED_STRING
I'm not a Coverity expert, but based on the documentation this could be something like:
``` C
/* coverity[ -tainted_string_sink_content : arg-0 ] */
void *
C_LoadModule(const char *mspec, CK_FUNCTION_LIST_PTR_PTR funcs)
{
```
What do you think?

frankmorgner:
I'm not an expert either, but I would avoid the annotation clutter in libp11's source code. I think its possible submit separate modeling files which covers stuff like this. Just have a look on Coverity's project site while being logged in...

PR #224, opened 2018-06-07
Improve Travis testing with different OpenSSL versions

ansasaki:
Create a matrix to test with different versions of OpenSSL. 

The matrix tests combinations of compiler (gcc or clang), OpenSSL version (1.0.2-stable or 1.1.0-stable), and host OS (linux).

The OpenSSL source is downloaded and built if not present.  The resulting binaries are stored in a cache to speed up new builds in Travis.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -43,8 +43,29 @@ sudo apt-get install -y libpcsclite-dev
 export CC=`which $CC`
 mkdir prerequisites
 cd prerequisites
+
+# Install OpenSSL if not present
+if [ -n "${OPENSSL}" ]; then
+    OPENSSL_DIR="${HOME}/openssl/${OPENSSL}"
+    if [ ! -f "${OPENSSL_DIR}/bin/openssl" ]; then
+        git clone https://github.com/openssl/openssl.git -b ${OPENSSL}
+        cd "openssl"
+        ./config shared -fPIC --prefix="${OPENSSL_DIR}"
+        make depend
+        make install
+        cd ..
+    fi
+
+    PATH="${OPENSSL_DIR}/bin:${PATH}"
+    CFLAGS="${CFLAGS} -I${OPENSSL_DIR}/include"
+    LDFLAGS="${LD_FLAGS} -L${OPENSSL_DIR}/lib"
+    LD_RUN_PATH="${OPENSSL_DIR}/lib"
+    LD_LIBRARY_PATH="${OPENSSL_DIR}/lib:${LD_LIBRARY_PATH}"
+fi
+
 install_from_github OpenSC OpenSC master
 # softhsm is required for "make check"
 install_from_github opendnssec SoftHSMv2 master --disable-gost
+
 cd ..
 rm -rf prerequisites
--- a/.travis.sh
+++ b/.travis.sh
@@ -23,17 +23,35 @@ set -e
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
 # THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+fetch_from_github() {
+    git clone https://github.com/$1/$2.git -b $3 --depth=1
+}
+
 install_from_github() {
     echo "Installing $2"
-    git clone https://github.com/$1/$2.git -b $3
+    fetch_from_github $1 $2 $3
     cd $2
     autoreconf -fvi
-    ./configure $4
+    ./configure $4 $5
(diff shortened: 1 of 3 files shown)


frankmorgner:
instead of updating the version number every time, you should use the git branches `OpenSSL_1_1_0-stable`, `OpenSSL_1_0_2-stable` and possibly more (see https://www.openssl.org/source/gitrepo.html). If you use https://github.com/openssl/openssl, you could even try the shortcut with `install_from_github`

ansasaki:
@frankmorgner Ok, changed the sources.

frankmorgner:
I think OpenSSL should be installed globally, just like the other dependencies (see `install_from_git` as reference).

frankmorgner:
Actually I don't care too much, but since we're talking about the testing system we have complete control over, I'd like to see more consistency (i.e. everything local or everything global).

ansasaki:
@frankmorgner Changed it to install OpenSSL globally and (try to) remove the existing pre-installed version.

Some restrictions made the final script not so different from the original one:

- OpenSSL don't have a ``configure`` script, but a ``config`` script. I could change the ``install_from_github()`` function, but I didn't
- In OpenSSL 1.0.2, ``config && make && sudo make install`` do not install the library properly, resulting in the previously existing OpenSSL to continue as the default version. Removing the previous OpenSSL version using ``apt-get remove``do not really remove the DSO. I believe some executable using OpenSSL is running when it tries to uninstall, resulting in the library to remain. And since the ``make install`` function do not set symbolic links properly, the applications end up finding the old version when they try to load the module. In the end, I had to keep exporting the environment variables to make the applications to find the right OpenSSL version.
- In OpenSSL 1.1.0. ``config && make && sudo make install`` works as expected.
- Removed the cache for OpenSSL binaries.

.travis.sh
```diff
@@ -37,14 +37,36 @@ install_from_github() {
...
+    cd "openssl"
+    ./config shared -fPIC --prefix=${OPENSSL_DIR}
+    make depend && make
+    sudo make install
+    cd ..
+    PATH="${OPENSSL_DIR}/bin:${PATH}"
+    CFLAGS="${CFLAGS} -I${OPENSSL_DIR}/include"
+    LDFLAGS="${LD_FLAGS} -L${OPENSSL_DIR}/lib"
+    LD_RUN_PATH="${OPENSSL_DIR}/lib"
+    LD_LIBRARY_PATH="${OPENSSL_DIR}/lib:${LD_LIBRARY_PATH}"
```
frankmorgner: Sorry for being nit-picky, but changing these variables should not be needed (here and in `.travis.yml`). please just use `sudo ldconfig`.

.travis.sh
```diff
@@ -37,14 +37,36 @@ install_from_github() {
...
 mkdir prerequisites
 cd prerequisites
+
+# Install OpenSSL
+if [ -n "${OPENSSL}" ]; then
+    # Remove pre-installed OpenSSL
+    sudo apt-get remove openssl libssl-dev
+
+    OPENSSL_DIR=/usr
+    git clone https://github.com/openssl/openssl.git -b ${OPENSSL}
```
frankmorgner: please add ` --depth=1`.

This optimization should also be used in `install_from_github`. `install_from_github` could even be split up into `install_from_github`, which calls `fetch_from_github`, where the latter can also be used for OpenSSL...

.travis.sh
```diff
@@ -37,14 +37,36 @@ install_from_github() {
...
+    cd "openssl"
+    ./config shared -fPIC --prefix=${OPENSSL_DIR}
+    make depend && make
+    sudo make install
```
frankmorgner: please use `make install_sw`

.travis.sh
ansasaki: I too wanted to get rid of these exports. Simply calling ``sudo ldconfig`` did not solve the issue. But following [this guideline](https://wiki.openssl.org/index.php/Compilation_and_Installation#PREFIX_and_OPENSSLDIR) to set the flags for both OpenSSL 1.0.2 and OpenSSL 1.1.0, and changing the prefix for OpenSSL helped.

The final issue was SoftHSM configuration not finding OpenSSL. Setting manually the prefix using ``--with-openssl`` did the trick.

.travis.sh
ansasaki: Applied the suggested changes in 75fa772

.travis.sh
ansasaki: Applied the suggested changes in 75fa772

frankmorgner:
Thanks! If you find some more time, you may want to also add libressl to the list...

Issue #234, opened 2018-07-31
memory leak at eng_back.c


defeden:
valgrind shows this leak

==26345== 7 bytes in 1 blocks are definitely lost in loss record 5 of 1,217
==26345==    at 0x4C2AB80: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==26345==    by 0x5199DF2: CRYPTO_malloc (in /lib/x86_64-linux-gnu/libcrypto.so.1.0.0)
==26345==    by 0x5199E4F: CRYPTO_strdup (in /lib/x86_64-linux-gnu/libcrypto.so.1.0.0)
==26345==    by 0x6A94AFC: parse_slot_id_string (eng_parse.c:200)
==26345==    by 0x6A92E7F: ctx_load_key (eng_back.c:647)
==26345==    by 0x6A940A9: ctx_load_privkey (eng_back.c:849)
==26345==    by 0x5212DCC: ENGINE_load_private_key (in /lib/x86_64-linux-gnu/libcrypto.so.1.0.0)

mouse07410:
> `==26345== by 0x5199E4F: CRYPTO_strdup (in /lib/x86_64-linux-gnu/libcrypto.so.1.0.0)`

It appears that the potential memory leak occurs in the OpenSSL code (can't tell where exactly, because OpenSSL wasn't compile in debug mode)?

dengert:
No, the error was the dupped string was not freed. 
Looks more like a problem in `ctx_load_key`. Although `ctx_load_key` will free `key_label` at line 836, there are many other `return NULL;` statements that do not free `key_label`. 
Maybe these `return NULL;` statements should be `goto error;` and `error:` inserted between line 834 and 835. 

Something like this compiled but untested patch:
[eng_back.c_key_label.diff.txt](https://github.com/OpenSC/libp11/files/2247769/eng_back.c_key_label.diff.txt)

mouse07410:
Thank you - your patch makes sense, and appears to work (tested with SoftHSMv2 and Yubikey NEO). I've merged it into [my fork](https://github.com/mouse07410/libp11.git). *Cannot tell whether it's sufficient, but it does seem necessary.*

defeden:
@dengert : your patch seems to be GOOD, thank you

ansasaki:
@mouse07410 @dengert Can you please open a PR with the proposed patch?

PR #235, opened 2018-08-02
Fix for memory leak at eng_back.c #234

dengert:
key_label needs to be freed . See Issue #234 for test results.

 Changes to be committed:
	modified:   src/eng_back.c


mtrojnar:
It looks good.  Thank you.

PR #236, opened 2018-08-13
add pin-source feature for pkcs11 uri

ortigali:
Implement "pin-source" attribute using below specs

https://tools.ietf.org/html/rfc7512#page-9

### DIFF
--- a/src/eng_parse.c
+++ b/src/eng_parse.c
@@ -32,11 +32,6 @@
 #include <string.h>
 #include <limits.h>
 
-#if defined(_WIN32) || defined(_WIN64)
-#define popen _popen
-#define pclose _pclose
-#endif
-
 static int hex_to_bin(ENGINE_CTX *ctx,
 		const char *in, unsigned char *out, size_t *outlen)
 {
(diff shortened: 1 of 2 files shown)


ortigali:
Hi, Could you review this PR?

mtrojnar:
I promise I will. Just give me a few days.

ortigali:
Thank you.

mtrojnar:
@frankmorgner Yes, `popen(3)` indeed introduces additional risks in applications where attackers may control environment variables, but not the name of the executable.  I don't think this is likely to be the case for our engine.  I think exposing the `pin-source` functionality may be dangerous in some applications regardless of the implementation.  This, however, is a vulnerability of the RFC, not specific to our implementation.
An obvious advantage of using `popen(3)` is potability.  Alternative solutions would require at least a separate Windows implementation.

mtrojnar:
@frankmorgner What I meant was not "I am too lazy to implement a proper solution", but rather "I think that a complex (and thus error-prone) solution to fix a problem that unlikely to occur would do more harm than good here".  This is just my opinion and I understand and appreciate your point of view.

ortigali:
I agree with @mtrojnar that 

> exposing the pin-source functionality may be dangerous in some applications regardless of the implementation

@frankmorgner Could you suggest an alternative solution for `popen`?

src/eng_parse.c
```diff
@@ -309,6 +310,22 @@ static int read_from_command(ENGINE_CTX *ctx,
 	return 1;
 }
 
+static int strnicmp(const char *cs, const char *ct, size_t count)
```
frankmorgner: have you tried `strncasecmp()`?

frankmorgner:
An alternative to `popen` would be to use `fork` and the `exec` family, as suggested by the man page. That, however, would require a seperate implementation for Windows (MSDN has [example code](https://docs.microsoft.com/en-us/windows/desktop/ProcThread/creating-a-child-process-with-redirected-input-and-output)). Indeed, that's more complex than just using `popen()`, but given the examples of both variants, I don't think that there's too much that can go wrong.

Anyway, if you still want to use `popen`, maybe it's possible to deactivate this feature by default. A note in the documentation for activating the feature can then make the preconditions for safe use more apparent.

frankmorgner:
This pull request **introduces 1 alert** when merging a659aba359fc136ca190f3f910d44babafe228df into 97700cb51ac1e84f5ac8bc402e6f9e0fc271d76b - [view on LGTM.com](https://lgtm.com/projects/g/OpenSC/libp11/rev/pr-cbe7045e24c2136efd9f6d720c7065a0d58d8b66)

**new alerts:**

* 1 for Wrong type of arguments to formatting function

---

*Comment posted by [LGTM.com](https://lgtm.com)*

src/eng_parse.c
```diff
@@ -263,6 +265,197 @@ static int parse_uri_attr(ENGINE_CTX *ctx,
 	return ret;
 }
 
+static int read_from_file(ENGINE_CTX *ctx,
+	const char *path, unsigned char *field, size_t *field_len)
+{
+	FILE *fp;
+
+	fp = fopen(path, "r");
```
frankmorgner: could use OpenSSL's bio methods

frankmorgner:
@ortigali an alternative to popen is suggested in the man page (linked above with http://manpages.ubuntu.com/manpages/trusty/en/man3/popen.3posix.html#rationale)

mtrojnar:
I merged #309 instead of this PR.

Issue #251, opened 2018-09-24
Certificate X509 mandatory attributes are not loaded


FdLSifu:
While reading PKCS11 2.40, a certificate CKC_X_509 supports following fieldd that are X509 mandatory :
- CKA_ISSUER
- CKA_SERIAL_NUMBER
- CKA_HASH_OF_SUBJECT_PUBLIC_KEY
- CKA_HASH_OF_ISSUER_PUBLIC_KEY
- CKA_NAME_HASH_ALGORITHM

CKA_NAME_HASH_ALGORITHM is security relevant cause if a certificate have been create using SHA-256, pkcs11 set the hash algorithm to default which is SHA-1. We know that SHA-1 is deprecated and due to missing attributes handling we decrease the hash length.

Should I present some code ?

mtrojnar:
Sure.  Please submit a PR.

mtrojnar:
Implemented in https://github.com/OpenSC/libp11/pull/252

PR #252, opened 2018-09-24
Store following certificate attributes: issuer, hash algorithm and hash of subject & public key

FdLSifu:
Pull-request associated to this issue: https://github.com/OpenSC/libp11/issues/251

### DIFF
--- a/src/p11_cert.c
+++ b/src/p11_cert.c
@@ -284,6 +284,10 @@ int pkcs11_store_certificate(PKCS11_TOKEN *token, X509 *x509, char *label,
 	CK_ATTRIBUTE attrs[32];
 	unsigned int n = 0;
 	int rv;
+    const EVP_MD* evp_md;
+    CK_MECHANISM_TYPE ckm_md;
+    unsigned char md[EVP_MAX_MD_SIZE];
+    unsigned int md_len;
 
 	/* First, make sure we have a session */
 	if (!PRIVSLOT(slot)->haveSession && PKCS11_open_session(slot, 1))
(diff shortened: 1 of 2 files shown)


src/p11_cert.c
```diff
@@ -295,6 +299,33 @@ int pkcs11_store_certificate(PKCS11_TOKEN *token, X509 *x509, char *label,
...
+        ckm_md = CKM_SHA256;
+    case NID_sha512:
+        ckm_md = CKM_SHA512;
+    case NID_sha384:
+        ckm_md = CKM_SHA384;
+    default:
+        ckm_md = 0;
+    }
+
+    /* Set hash algorithm; default is SHA-1 */
```
frankmorgner: if SHA-1 is the default, then please make that explicit in the switch statement above, i.e. move `default:` above `case NID_sha1:`.

FdLSifu: Actually SHA-1 is the default CKA_NAME_HASH_ALGORITHM.
Meaning that if ckm_md is equal to 0, SHA-1 will be use. This comes from PKCS11.

FdLSifu:
Done as requested

frankmorgner:
Your switch statement misses the `break;`. Have you ever tested this?

FdLSifu:
I have tested this by using function pkcs11_md2ckm from p11_pkey.c.
I have then copy back the code and you are right, break are missing.

mtrojnar:
@frankmorgner Are you satisfied with the implementation of changes you requested?

frankmorgner:
The code itself looks good, but I did not check the functionality with the PKCS#11 spec.

Issue #259, opened 2018-10-22
Enable signing with EC key


alex-nitrokey:
It would be great to have the [function pkcs11_pkey_ec_sign](https://github.com/OpenSC/libp11/blob/87a9176c817408da8baccef0c4e44ad6070f07c2/src/p11_pkey.c#L523) actually implemented. This would enable user to use a ECC generated on a card as CA key.

mtrojnar:
Implemented by #260.

PR #260, opened 2018-11-12
Add EC signing through EVP api

bryan-hunt:
Updated p11_pkey.c to implement pkcs11_try_pkey_ec_sign so EC keys can be used in various OpenSSL 1.1 use cases (server key, csr creation, client cert, etc) as OpenSSL 1.1 uses the EVP interface over the 1.0.2 EC interfaces..

Also reported as issue #259

Notes:
* An attempt was made to maintain the existing style and formatting. 
* All copyright claims to the changes cited in this PR are relinquished to the OpenSC project

### DIFF
--- a/src/libp11-int.h
+++ b/src/libp11-int.h
@@ -154,6 +154,7 @@ void CRYPTO_THREAD_lock_free(int);
 /* Emulate the OpenSSL 1.1 getters */
 #if OPENSSL_VERSION_NUMBER < 0x10100003L || defined(LIBRESSL_VERSION_NUMBER)
 #define EVP_PKEY_get0_RSA(key) ((key)->pkey.rsa)
+#define EVP_PKEY_get0_EC_KEY(key) ((key)->pkey.ec)
 #endif
 
 /* Reinitializing the module afer fork (if detected) */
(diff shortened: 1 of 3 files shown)


dwmw2:
See https://github.com/openssl/openssl/pull/7408 and make sure OpenSSL isn't going to ask you to use digests you don't support.

bryan-hunt:
@dwmw2 I'm not quite following the issue actually. Other than shorter digests (which is checked) there shouldn't be an issue. Per NIST ECDSA can sign any length of digest larger than it's bitlength you are just bound by the bits of security of the signing key. OpenSSL should be verifying the cipher suite however if the user has specified suite-b for example.

bryan-hunt:
I meant if a user specified a cipher suite OpenSSL shouldn't negotiate outside of it. Suite B is much more limited set of possibilities that would prohibit using the wrong digesting algorithm.

dwmw2:
Although the work referenced above was inspired by TPMv2 ENGINE requirements, we have also seen cases of PKCS#11 provider modules which *(inexplicably)* fail to perform ECDSA signatures on SHA512 signatures, and only work when TLS negotiates SHA256 or smaller.

mtrojnar:
Nice work.  Would it be possible to also provide a test case for the new feature?

PR #268, opened 2018-12-17
fix signed/unsigned and possible buffer overflow

melanger:
@gimlly @mgrabovsky @petrs


mtrojnar:
I couldn't find any unsafe use of this variable (such as greater/less comparisons), so I don't think there is any security impact.  Anyway, this fix looks like a nice code style improvement.

PR #271, opened 2019-01-25
fixed warnings from Clang Checker

frankmorgner:


frankmorgner:
I guess not. rc is never used as return code in auth.c and decrypt.c. But with getrandom.c it seems you're right. I'll double check.

frankmorgner:
I've restored the handling of `rc` of `getrandom.c`, but as said above, I've kept the semantics that auth.c and decrypt.c still return 1 in case of an error.

frankmorgner:
A thing that needs more discussion than some example program's return value, is the fact that clang's checker reported the possibility that a private key's `_private` member may be `NULL` (see my change in `p11_key.c`). In generak, I think libp11 could do better in terms of checking the internal structures. For example, `PRIVKEY`, `KEY2TOKEN`, `PRIVCERT` or `CERT2TOKEN` are always dereferencing pointers without checking their validity. I don't have enough insights to decide whether checking for `NULL` makes sense or not, so I'd prefer you to take a deeper look.

mtrojnar:
I'm re-implementing the changes proposed in this PR with other related fixes.  I'll close the PR when all the indicated issues are dealt with.

frankmorgner:
@mtrojnar please also mind the fix of `src/p11_key.c` from this PR.

mtrojnar:
@frankmorgner I did it 2 weeks ago: https://github.com/OpenSC/libp11/commit/d3a29acab78902a7e9bdea9ea1df316aefc52697

PR #274, opened 2019-02-11
Load EC public key from the certificate if public key object is not available

ansasaki:
If an EC private key is stored in a device, but the corresponding public key object is not, calling ENGINE_load_private_key() will return an incomplete EVP_PKEY structure, without the public key information, which will make some functions to fail (e.g. X509_check_private_key()).

This PR makes the engine to try, as a last resort, to load the public key from a matching certificate object.

### DIFF
--- a/src/p11_ec.c
+++ b/src/p11_ec.c
@@ -235,6 +235,10 @@ static int pkcs11_get_point(EC_KEY *ec, PKCS11_KEY *key)
 static EC_KEY *pkcs11_get_ec(PKCS11_KEY *key)
 {
 	EC_KEY *ec;
+	PKCS11_CERT *cert;
+	EVP_PKEY *pubkey;
+	EC_KEY *pubkey_ec;
+	const EC_POINT *point;
 	int no_params, no_point;
 
 	ec = EC_KEY_new();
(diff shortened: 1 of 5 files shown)


dwmw2:
cf. https://github.com/openssl/openssl/issues/1532 and https://github.com/openssl/openssl/pull/1551

src/p11_ec.c
```diff
@@ -251,6 +255,30 @@ static EC_KEY *pkcs11_get_ec(PKCS11_KEY *key)
 	if (no_point && key->isPrivate) /* Retry with the public key */
 		no_point = pkcs11_get_point(ec, pkcs11_find_key_from_key(key));
 
+	/* Retry with the certificate */
+	if (no_point && key->isPrivate) {
+		cert = pkcs11_find_certificate(key);
```
dwmw2: This returns a certificate which has the same CKA_ID as the key, doesn't it? In an ideal world that might match... but we shouldn't assume that. By all means look for such a cert as a best effort, but let's actually do a test operation and *check* that it matches, before filling it in.

Fixing OpenSSL would also be useful.

ansasaki: > This returns a certificate which has the same CKA_ID as the key, doesn't it? In an ideal world that might match... but we shouldn't assume that. By all means look for such a cert as a best effort, but let's actually do a test operation and _check_ that it matches, before filling it in.

Yes, it gets a certificate with matching CKA_ID. I though this would be acceptable since the same is done when searching for matching public key objects. As you pointed out, this is a best effort to improve the current situation.

If we choose to check if the public keys match before filling the structure, we should do the same for public key objects, and for other algorithms.

> 
> Fixing OpenSSL would also be useful.

After reading about your last attempt to fix it there, I feel discouraged to do so.

ansasaki: @dwmw2 Do you have a suggestion on how to check if the private key and a public key match for EC keys without performing a signature/verification?

For most of devices, doing an extra signature every time a key is used is impractical, since it can take too much time. I would suggest to check if the groups of the private key and the public key match and if the public key (point) is on the curve. But that would only guarantee that both  are using the same parameters.

dengert:
It is user's responsibility to provide matching cert, pubkey and privkey. 99.99% this will be correct. 
Within PKCS\#11 the CKA_ID is designed to link the cert, pubkey and privkey.
 
I agree with @ansasaki  who said: "doing an extra signature every time a key is used is impractical" 
Doing an extra signature with the private key may actually require the user to enter the PIN again. 
(Tokens with keys where "PIN Always"/user_consent/CKA_ALWAYS_AUTHENTICATE is needed.) 

The real problem is PKCS\#11 and OpenSSL do not agree on what is actually required to be visible in a private key.

dwmw2:
> The real problem is PKCS#11 and OpenSSL do not agree on what is actually required to be visible in a private key.

OpenSSL doesn't even agree with itself. Let's fix it.

mouse07410:
```
 CC       check-privkey.o
check-privkey.c:79:9: warning: implicit declaration of function 'PEM_read_X509'
      is invalid in C99 [-Wimplicit-function-declaration]
        cert = PEM_read_X509(cert_fp, NULL, NULL, NULL);
               ^
check-privkey.c:79:7: warning: incompatible integer to pointer conversion
      assigning to 'X509 *' (aka 'struct x509_st *') from 'int'
      [-Wint-conversion]
        cert = PEM_read_X509(cert_fp, NULL, NULL, NULL);
             ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2 warnings generated.
```
Is the above not a concern?

dengert:
check-privkey.c needs:
#include <openssl/pem.h>

mouse07410:
Yes. And p11_rc.c needs to remember that EC_KEY_set_public_key() returns 1 for success and 0 for error. Currently no_point is set to the wrong value.

PR #278, opened 2019-03-05
update for LibreSSL-2.8.3 compat to p11_rsa.c

patchMonkey156:
As LibreSSL implements TLS 1.3, more such methods will already be included. Their stack seems to avoid declaring things as "static", which causes conflict when OpenSSL-1.0.X compatibility methods are left in for developing and future versions of LibreSSL.

Not much to do but patch these concerns as they arise.

### DIFF
--- a/src/p11_rsa.c
+++ b/src/p11_rsa.c
@@ -402,7 +402,7 @@ static void free_rsa_ex_index()
 #endif
 }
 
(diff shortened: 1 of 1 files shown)


patchMonkey156:
Sorry about that last one!
I am sorta new to this, but very interested in contributing towards the project's success.

patchMonkey156:
Sorry about that last one!
I am sorta new to this, but very interested in contributing towards the project's success.

PR #283, opened 2019-04-10
add needed include for getpid()

HRio:
Fixes:
```
 p11_atfork.c: In function '_P11_get_forkid':
 p11_atfork.c:78:9: warning: implicit declaration of function 'getpid'; did you mean 'getenv'? [-Wimplicit-function-declaration]
  return getpid();
```

### DIFF
--- a/src/p11_atfork.c
+++ b/src/p11_atfork.c
@@ -23,6 +23,7 @@
 #include "libp11-int.h"
 
 #ifndef _WIN32
+#include <unistd.h>
 
 #ifndef __STDC_VERSION__
 /* older than C90 */


src/p11_atfork.c
```diff
@@ -21,6 +21,7 @@
  */
 
 #include "libp11-int.h"
+#include <unistd.h>
```
HRio: should I add `#ifndef _WIN32` around?
https://ci.appveyor.com/project/LudovicRousseau/libp11/builds/23751293/job/2o1aangcodd28lg3

not sure what the WIN32 equivalent header file would be...

PR #292, opened 2019-06-06
Make ECDH derived keys extractable to support 'openssl pkeyutl -derive'.

aasimon:
Change has been tested with openssl using SoftHSMv2 engine.

### DIFF
--- a/src/p11_ec.c
+++ b/src/p11_ec.c
@@ -558,6 +558,8 @@ static int pkcs11_ecdh_derive(unsigned char **out, size_t *outlen,
 		{CKA_TOKEN, &false, sizeof(false)}, /* session only object */
 		{CKA_CLASS, &newkey_class, sizeof(newkey_class)},
 		{CKA_KEY_TYPE, &newkey_type, sizeof(newkey_type)},
+		{CKA_SENSITIVE, &false, sizeof(false) },
+		{CKA_EXTRACTABLE, &true, sizeof(true) },
 		{CKA_ENCRYPT, &true, sizeof(true)},
 		{CKA_DECRYPT, &true, sizeof(true)}
 	};
@@ -581,7 +583,7 @@ static int pkcs11_ecdh_derive(unsigned char **out, size_t *outlen,
 			return -1;
 	}
 
-	rv = CRYPTOKI_call(ctx, C_DeriveKey(spriv->session, &mechanism, kpriv->object, newkey_template, 5, &newkey));
+	rv = CRYPTOKI_call(ctx, C_DeriveKey(spriv->session, &mechanism, kpriv->object, newkey_template, sizeof(newkey_template)/sizeof(*newkey_template), &newkey));
 	CRYPTOKI_checkerr(CKR_F_PKCS11_ECDH_DERIVE, rv);
 
 	/* Return the value of the secret key and/or the object handle of the secret key */


mtrojnar:
The very purpose of generating keys on a hardware engine is to make them **not** extractable.  Otherwise it's way faster to generate the keys on the main CPU and then write them to the engine.

aasimon:
I am not extracting the ECC key itself, but only the ECDH derived AES key.
In my particular use-case I need to perform the AES decryption on the CPU because the interface to the crypto-chip is too slow to handle the data-rate.

mtrojnar:
The obvious solution is to use the derived AES key as a symmetric key wrapping key, and only use it to protect distributing your actual symmetric data encryption key.
https://en.wikipedia.org/wiki/Cryptographic_key_types

dengert:
The problem is more complicated. It depends on what the token can do and what the PKCS11 module can do with different tokens. 

 For example PIV cards return the derived key as the response to the GENERAL AUTHENTICATE ADPU command. This is described in [NIST sp800-73-4](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73-4.pdf)  as an example in Part 2, Appendix A.5.2 "Elliptic Curve Cryptography Diffie-Hellman" and A.5.2.1.1 "The GENERAL AUTHENTICATE Command". "A.5.2.2 PIV KMK Specific ECDH Key Agreement Schemes" describes how this is used in "C(2e, 2s) – Each party has a static key pair and generates an ephemeral key pair [SP800-56A, Section 6.1.1].

So PIV returns a secret key which is stored in effect in a session object in software as CKA_TOKEN=FALSE, CKA_SENSITIVE=FALSE, CKA_EXTRACTABLE=TRUE and  CK_KEY_TYPE=CKK_GENERIC_SECRET.

The PIV token does not support using a derived key on the token.

Other tokens such as the  SoftHSMv2 may have the ability to do both and not clear what is the default, which may depend on what PKCS11 module is being used.   The token may even  have the ability to generate an ephemeral key pair on the token. (If you only use the two static keys, you will end up with the same secret key each time. Not a good idea.)  It is not clear how `openssl pkeyutl -derive` is being used but it looks like the key needs to be returned in any case. 

So a more general solution for libp11 might need to include some additional parameters or not use openssl pkeyutl -derive.

mtrojnar:
> The PIV token does not support using a derived key on the token

@dengert Do you mean that CK_KEY_TYPE=CKK_GENERIC_SECRET only allows for using the computed shared secret for deriving an actual encryption key or message authentication secret, or that CKA_TOKEN=FALSE only makes it only available for the current session? How do you suggest to address it?

Honestly, I don't think that support for `openssl pkeyutl -derive` is an essential requirement.

mouse07410:
> I don't think that support for `openssl pkeyutl -derive` is an essential requirement.

I think being able to derive (and use!) keys *is* an essential requirement. But I don't understand what the problem is with the current code base (tested with Yubikey 4):
```
$ openssl pkeyutl -engine pkcs11 -keyform ENGINE -derive -peerform PEM -inkey "pkcs11:model=PKCS%2315%20emulated;manufacturer=piv_II;serial=xxxxxx;token=XXX;id=%03;object=KEY%20MAN%20key;type=private" -peerkey /tmp/ec384-pub.pem -hexdump
engine "pkcs11" set.
GOST engine already loaded
Enter PKCS#11 token PIN for XXX:
0000 - 14 eb 8a bb 74 9b 34 3c-32 97 13 cb a0 1b 57 fe   ....t.4<2.....W.
0010 - 41 4c ae 80 2b 0a 2c 50-1c 46 6d df c2 03 97 13   AL..+.,P.Fm.....
0020 - dd 53 71 36 af cd e0 97-b3 bc 38 66 70 b0 ed b0   .Sq6......8fp...
$ openssl pkeyutl -engine pkcs11 -peerform ENGINE -derive -keyform PEM -peerkey "pkcs11:model=PKCS%2315%20emulated;manufacturer=piv_II;serial=xxxxxxx;token=XXX;id=%03;object=KEY%20MAN%20pubkey;type=public" -inkey /tmp/ec384-priv.pem -hexdump
engine "pkcs11" set.
GOST engine already loaded
0000 - 14 eb 8a bb 74 9b 34 3c-32 97 13 cb a0 1b 57 fe   ....t.4<2.....W.
0010 - 41 4c ae 80 2b 0a 2c 50-1c 46 6d df c2 03 97 13   AL..+.,P.Fm.....
0020 - dd 53 71 36 af cd e0 97-b3 bc 38 66 70 b0 ed b0   .Sq6......8fp...
$ 
```

Having said that - I don't see any harm merging this PR, as everything still works (so it does not appear to break things):
```
$ openssl3 pkeyutl -engine pkcs11 -keyform ENGINE -derive -peerform PEM -inkey "pkcs11:model=PKCS%2315%20emulated;manufacturer=piv_II;serial=cxxxxxxx;token=XXX;id=%03;object=KEY%20MAN%20key;type=private" -peerkey /tmp/ec384-pub.pem -hexdump
engine "pkcs11" set.
Enter PKCS#11 token PIN for XXX:
0000 - 14 eb 8a bb 74 9b 34 3c-32 97 13 cb a0 1b 57 fe   ....t.4<2.....W.
0010 - 41 4c ae 80 2b 0a 2c 50-1c 46 6d df c2 03 97 13   AL..+.,P.Fm.....
0020 - dd 53 71 36 af cd e0 97-b3 bc 38 66 70 b0 ed b0   .Sq6......8fp...
$ openssl3 pkeyutl -engine pkcs11 -peerform ENGINE -derive -keyform PEM -peerkey "pkcs11:model=PKCS%2315%20emulated;manufacturer=piv_II;serial=xxxxxxx;token=XXX;id=%03;object=KEY%20MAN%20pubkey;type=public" -inkey /tmp/ec384-priv.pem -hexdump
engine "pkcs11" set.
0000 - 14 eb 8a bb 74 9b 34 3c-32 97 13 cb a0 1b 57 fe   ....t.4<2.....W.
0010 - 41 4c ae 80 2b 0a 2c 50-1c 46 6d df c2 03 97 13   AL..+.,P.Fm.....
0020 - dd 53 71 36 af cd e0 97-b3 bc 38 66 70 b0 ed b0   .Sq6......8fp...
$ 
```
Tested with OpenSSL-1.1.1c and 3.0-dev.

dengert:
In response to https://github.com/OpenSC/libp11/pull/292#issuecomment-527889196

The basic output of a ECDH derive operation is a CKK_GENERIC_SECRET. 

2.3.9 Elliptic curve Diffie-Hellman key derivation
"This mechanism derives a secret value, and truncates the result according to the CKA_KEY_TYPE
attribute of the template and, if it has one and the key type supports it, the CKA_VALUE_LEN attribute of the template. (The truncation removes bytes from the leading end of the secret value.) The mechanism contributes the result as the CKA_VALUE attribute of the new key; other attributes required by the key type must be specified in the template."

So it depends on what the CK_EC_KDF_TYPE is supported by the token or in software and what the  CKA_KEY_TYPE in the template wants. 

2.3.8 EC mechanism parameters
"CK_EC_KDF_TYPE is used to indicate the Key Derivation Function (KDF) applied to derive keying data
from a shared secret. The key derivation function will be used by the EC key agreement schemes. It is
defined as follows."

The PIV in hardware only supports CKD_NULL. Other tokens may support functions in hardware. 
Software may support using the  CKK_GENERIC_SECRET to produce an AES key by: "The truncation removes bytes from the leading end of the secret value." Depends on the PKCS\#11 implementation.

The OpenSC pkcs11-tool has a example to return a secret key.

dengert:
When deriving an AES key you may need to add to template 
`{ CKA_VALUE, NULL, 256 }`
Or what ever size is needed.

mtrojnar:
On second thought, setting CKA_SENSITIVE=FALSE and CKA_EXTRACTABLE=TRUE seems harmless as long as it's used with CKA_TOKEN=FALSE (i.e., not persistent).  I'm merging this PR.

PR #305, opened 2019-09-26
Update INSTALL.md

maxcuttins:
Addeded the needed to have `autoconf ` pkg and fixed the format of bash commands


mtrojnar:
This is not needed for building release tarballs.

PR #324, opened 2020-01-09
Handle keys without label attribute

efternavn:


dwmw2:
If a label is specified in the URI but not present on a given object, that object shouldn't match.

efternavn:
@dwmw2, you mean if both Label and ID are specified?
Then a key without label is handled the same way as a key with a non-matching label — meaning that if the IDs match, the key matches — as far as I can see.
I didn’t want to change that behavior even though I think it is a bit strange, but maybe there’s something I misunderstood in the code.

mtrojnar:
@efternavn I merged it. Next time, when submitting your code to an existing project, please try to retain the original coding style even if you think it's excessively verbose or redundant.  You may then propose changing the coding style as a separate PR.  One think that is even worse than bad coding style is **inconsistent** coding style...

efternavn:
Thanks @mtrojnar.
And I completely agree. You're referring to the removal of `!= NULL` in conditions, right? I found that the code was already inconsistent -- for instance with a line like this `if (s_slot_cert_id && *s_slot_cert_id)`.
But I'll keep it in mind.

mtrojnar:
> You're referring to the removal of `!= NULL` in conditions, right?

Right.

> I found that the code was already inconsistent

Would you like to submit a PR to clean this up?

efternavn:
> > I found that the code was already inconsistent
> 
> Would you like to submit a PR to clean this up?

I'll have a look to get an idea of how big a task it would be.
Where should we discuss which style is preferred? Would you like me to create a new issue?
The C++ Core Guidelines recommend the short form: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es87-dont-add-redundant--or--to-conditions

dwmw2:
Indeed, getting rid of all that utterly redundant  `== 0` would be a good thing.

I never understood why people would write `if ( something != 0 )` in the first place. And if they misunderstand C truth values in that way, why they don't get recursive and end up with `if ((something != 0) != 0)` then `if (((something != 0) != 0) != 0)`...

mtrojnar:
I like the idea of using `if (p)` and `if (!p)`.  `p != NULL` is almost as ugly as `return (value);`.
Yes, please open a new issue if you need to discuss something else.

Issue #233, opened 2018-07-17
Re-license files under a known license


ansasaki:
Some files are not licensed under a known permissive license, making it difficult to be used and distributed. For instance, the following files:

https://github.com/OpenSC/libp11/blob/master/examples/auth.c
https://github.com/OpenSC/libp11/blob/master/examples/listkeys.c
https://github.com/OpenSC/libp11/blob/master/examples/decrypt.c
https://github.com/OpenSC/libp11/blob/master/examples/getrandom.c
https://github.com/OpenSC/libp11/blob/master/tests/fork-test.c
https://github.com/OpenSC/libp11/blob/master/tests/list-tokens.c
https://github.com/OpenSC/libp11/blob/master/tests/openssl_version.c

Please, re-license the files using a known license. This would require the original authors to agree with the new licensing. One good licensing option would be the BSD 2-Clause, which is already used in some files.

mtrojnar:
Please note the listed files are tests and examples that probably won't be included in binary distributions.  Source distributions may however be affected.  I hereby agree to re-license my contributions in those 7 files as BSD 2-Clause.

ansasaki:
I would like to invite the original authors to please agree in this re-licensing:
@nmav 
@frankmorgner 
@szikora 

I could not find a github for Andreas Jellinghaus. Maybe someone could please contact him?

nmav:
I'm fine with using BSD-2 clause license for my contributions.

szikora:
Hi,

It’s fine for me to modify the licensing terms to any BSD type licence!

Cheers,

Jean-Pierre

mtrojnar:
@ansasaki It seems that you have collected permissions from from all the identified authors.  Please feel free to submit a PR with license updates.

nased0:
I am fine with BSD license too!

mtrojnar:
No PR has been submitted for months.  Apparently nobody is interested in re-licensing the code anymore.

PR #326, opened 2020-01-31
Re-license source code files under a known license

ansasaki:
Some source code files weren't licensed properly.  This adds BSD style
licensing to them as agreed in:

https://github.com/OpenSC/libp11/issues/233

Some wording in some files were changed to avoid misleading the reader.

### DIFF
--- a/examples/README
+++ b/examples/README
@@ -2,8 +2,6 @@ Libp11 example code
 ===================
 
 This directory contains some example code how to use libp11.
-Feel free to use this code in any way, it is public domain,
-not copyrighted.
 
 auth.c		Example for authentication, i.e. get the first
 		token, get the first certificate, ask for pin,
--- a/examples/auth.c
+++ b/examples/auth.c
@@ -1,10 +1,39 @@
+/*
+ * Copyright © 2020, Andreas Jellinghaus <andreas@ionisiert.de>
+ * All rights reserved.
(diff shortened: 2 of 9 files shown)


mtrojnar:
Why did you choose 3-clause BSD over 2-caluse BSD?
https://en.wikipedia.org/wiki/BSD_licenses
https://mail-index.netbsd.org/netbsd-announce/2008/06/20/msg000030.html

ansasaki:
> Why did you choose 3-clause BSD over 2-caluse BSD?
> https://en.wikipedia.org/wiki/BSD_licenses
> https://mail-index.netbsd.org/netbsd-announce/2008/06/20/msg000030.html

Because Andreas Jellinghaus agreed on 3-clause BSD (https://github.com/OpenSC/libp11/issues/233#issuecomment-410626994).

I can change it to 2-clause BSD if he, as the others, agree.

mtrojnar:
On one hand it is good for examples be public domain to promote code reuse.  On the other hand coding style of our examples is so bad that I don't oppose to using a more restrictive license.

I still believe 2-clause BSD is better (less restrictions, more freedom), but 3-clause BSD is good enough for the code that got re-licensed.

mtrojnar:
@ansasaki Thank you for your work making this PR.

PR #333, opened 2020-04-03
support pin-source with piping from a subprocess

keldonin:
As specified in [RFC7512 section 2.4](https://tools.ietf.org/html/rfc7512#section-2.4):
```
o  If the value contains "|<absolute-command-path>", the
      implementation SHOULD read the PIN from the output of an
      application specified with absolute path "<absolute-command-
      path>".  Note that character "|" representing a pipe does not have
      to be percent-encoded in the query component of a PKCS #11 URI.
```

This PR maybe a naive implementation, happy to make it right for inclusion.

### DIFF
--- a/src/eng_parse.c
+++ b/src/eng_parse.c
@@ -287,6 +287,54 @@ static int read_from_file(ENGINE_CTX *ctx,
 	return 1;
 }
 
+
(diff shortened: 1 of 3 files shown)


keldonin:
Saw comment (too late!) on #236 and #309, withdrawing for now until secure implementation is available.

PR #350, opened 2020-09-10
Fix potential leak in RSA method

kFYatek:
Some OpenSSL functions (e.g. `pkey_rsa_encrypt()`) may cache some information inside an `RSA` object. See also `rsa_ossl_finish()` in OpenSSL sources.

libp11 has not been properly freeing those, causing a memory leak. This might be a problem if libp11 is registered as the default RSA method (`ENGINE_register_RSA()`, `ENGINE_set_default_RSA()`), which is especially possible if e.g. `OPENSSL_init_crypto(OPENSSL_INIT_ENGINE_ALL_BUILTIN, NULL)` has been called, like SoftHSM does.

### DIFF
--- a/src/p11_rsa.c
+++ b/src/p11_rsa.c
@@ -352,6 +352,11 @@ int (*RSA_meth_get_priv_dec(const RSA_METHOD *meth))
     return meth->rsa_priv_dec;
 }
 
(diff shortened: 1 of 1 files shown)


mtrojnar:
Merged in #352.

kFYatek:
Um, sorry, but it looks that you haven't actually merged changes from this PR. #352 was a separate issue.

PR #352, opened 2020-09-10
Fix potential leak in pkcs11_try_pkey_ec_sign()

kFYatek:
`ossl_sig` was not properly freed if any of the checks below it failed. This was especially likely for the `check_key_fork(key)` check, if the key loaded from outside the engine was used.

### DIFF
--- a/src/p11_pkey.c
+++ b/src/p11_pkey.c
@@ -542,7 +542,7 @@ static int pkcs11_try_pkey_ec_sign(EVP_PKEY_CTX *evp_pkey_ctx,
 	EVP_PKEY *pkey;
 	EC_KEY *eckey;
 	PKCS11_KEY *key;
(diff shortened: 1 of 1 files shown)


src/p11_pkey.c
```diff
@@ -542,7 +542,7 @@ static int pkcs11_try_pkey_ec_sign(EVP_PKEY_CTX *evp_pkey_ctx,
 	EVP_PKEY *pkey;
 	EC_KEY *eckey;
 	PKCS11_KEY *key;
-	int rv = 0;
+	int rv = -1;
```
mtrojnar, line 545: What about `CKR_GENERAL_ERROR`?

kFYatek: The specific value is discarded anyway, but sure, why not.

PR #355, opened 2020-09-23
UI_dup_input_string() returns <= 0 on error

botovq:
According to [OpenSSL's documentation][1], the return values of most
UI_add_ and UI_dup_ functions are "a positive number on success or
a value which is less than or equal to 0 otherwise."

This is also readily seen from [the source][2] as `general_allocate_string()`
will typically return `-1`.

[1]: https://www.openssl.org/docs/man1.1.1/man3/UI_dup_input_boolean.html#RETURN-VALUES
[2]: https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/ui/ui_lib.c#L206-L223

### DIFF
--- a/src/eng_back.c
+++ b/src/eng_back.c
@@ -135,8 +135,8 @@ static int ctx_get_pin(ENGINE_CTX *ctx, const char* token_label, UI_METHOD *ui_m
 	if (!prompt) {
 		return 0;
 	}
-	if (!UI_dup_input_string(ui, prompt,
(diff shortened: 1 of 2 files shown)


botovq:
The appveyor builds failed due to unavailability of the `https://slproweb.com/download/Win32OpenSSL-1_1_1g.exe`. This should be updated to `https://slproweb.com/download/Win32OpenSSL-1_1_1h.exe` which was released yesterday.

PR #360, opened 2020-10-07
Fix potential leak in RSA method

kFYatek:
This is basically a re-open of #350 because @mtrojnar closed that PR but it does not seem to actually be addressed on master.


mtrojnar:
Yeah, I can no longer reopen the original PR after you created this one...

kFYatek:
Sorry about the mess. I wasn't sure if you'll see the commend on that closed PR, and didn't really know how to proceed.

mtrojnar:
Good point.  We do not override the default `init` method, so it definitely also makes sense to invoke the default `finish` method.

PR #369, opened 2020-11-03
Ecdsa segfault

kFYatek:
`EVP_get_digestbynid(signature_nid)` can return `NULL` for some signature types - for example when `signature_nid` is `NID_ecdsa_with_SHA256` - quite common case nowadays. This was causing a segmentation fault.

I added a call to `OBJ_find_sigid_algs()` and some additional precautions to extract the actual hash algorithm NID first.

I also added a test for `PKCS11_store_certificate()` that crashes on `master`.


mtrojnar:
Nice PR.  Thank you.

PR #376, opened 2020-11-27
clean up OpenSSL 1.1.0 compatibility declarations on RSA struct and methods

DDvO:
When building with OpenSSL 1.1.1d, I noticed that the lib code contains clutter due to OpenSSL version compatibility workarounds.
I moved those on accessing the RSA structure and methods into `src/libp11-int.h`.

### DIFF
--- a/src/libp11-int.h
+++ b/src/libp11-int.h
@@ -29,6 +29,112 @@
 #define CRYPTOKI_EXPORTS
 #include "pkcs11.h"
 
+#if OPENSSL_VERSION_NUMBER < 0x10100005L || defined(LIBRESSL_VERSION_NUMBER)
+#include <openssl/dso.h>
+inline const BIGNUM *RSA_get0_n(const RSA *r)
+{
+    return r->n;
+}
+inline const BIGNUM *RSA_get0_e(const RSA *r)
+{
+    return r->e;
+}
+inline const BIGNUM *RSA_get0_d(const RSA *r)
+{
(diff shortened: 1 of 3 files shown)


mtrojnar:
1. Based on https://travis-ci.org/github/OpenSC/libp11/builds/746217669 this PR fails with OpenSSL versions other than 1.1.1.
2. Inline functions require C99 (https://en.wikipedia.org/wiki/Inline_function).  Should we allow older dialects?
3. I'm not convinced that using inline functions improves code quality.

dengert:
An other approach you may what to consider is using what OpenSC does since 2016. https://github.com/OpenSC/OpenSC/blob/master/src/libopensc/sc-ossl-compat.h Most code in OpenSC is based on OpenSSL API 1.1.  The  macros in sc-ossl-compat.h define missing handlers if used with older versions of OpenSSL or LibreSSL. 
As these older versions are no longer used eventual there will be nothing left for  sc-ossl-compat.h. to define.
There is even a change from 2019 for OpenSSL-3.0.0 for ECC.

There is a test to see if the compiler can use inline, taken from OpenSSL-1.1.0 e_os2.h. Having these inline routines in a header file also means that they are only compiled if the source code calls the routine. 

It may not have all the macros or inline routines you need, but they could be added. It has greatly simplified the OpenSC code and works on every platform OpenSC is built on. Linux, Windows and MacOS and maybe others.

Issue #377, opened 2020-12-02
PKCS11_generate_key DEPRECATED after change to call C_GenerateKeyPair


williamcroberts:
PR #188 changes PKCS11_generate_key to call C_GenerateKeyPair() instead of generating the keys in software and storing them to the HSM. It removes the DEPRECATED warning from PKCS11_generate_key.

However in https://github.com/OpenSC/libp11/commit/c1c2753176421a148d02296a893a2fbe6374172f#diff-887ce845c19283ef442b0ee4b98a686c73a13d5988b452f0577dd96f6c48f876R439 the routine becomes re-marked as deprecated, was that intentional?

PR #378, opened 2020-12-02
PKCS11_generate_key remove from DEPRECATED

williamcroberts:
PKCS11_generate_key was removed from the DEPRECATED list when the
internal implementation changed from generating the key material in
software and storing to the HSM to generating in the HSM directly
through the PKCS#11 call C_GenerateKeyPair.

However, commit:
  - c1c275317642 C89 fixes and style unification

Added that routine back to the DEPRECATED list even though the internal
implementation was corrected. This appears to be some form of a merge
bug.

Fixes: #377

Signed-off-by: William Roberts <william.c.roberts@intel.com>

### DIFF
--- a/src/libp11.h
+++ b/src/libp11.h
@@ -356,6 +356,22 @@ extern int PKCS11_store_private_key(PKCS11_TOKEN * token, EVP_PKEY * pk, char *l
 extern int PKCS11_store_public_key(PKCS11_TOKEN * token, EVP_PKEY * pk, char *label, unsigned char *id, size_t id_len);
 
 /**
+ * Generate and store a private key on the token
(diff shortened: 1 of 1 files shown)


mtrojnar:
The idea behind deprecating this function was to replace it with `EVP_PKEY_keygen()` (see the engine example in the manual).  Basically, we should register our own `EVP_PKEY_meth_set_keygen()` method in `p11_pkey.c`, but we haven't implemented it yet.
The advantage of the proposed approach is that such interface also allows for generating EC and not only RSA keys.

mtrojnar:
Sorry.  I didn't mean to merge it.

PR #380, opened 2020-12-23
if pkcs11 URI consists of 'slot-id', skip the slot searching algorithm.

YetiFromTheAlps:
Hi there!

Recently I had some problems on referencing the right key using an URI, which was stored on only one of multiple tokens, all of the same sort. Unfortunately for some other reason I could not use the serial ID, so the only attribute that could identify the correct token was the slot.

This is a proposal to support 'slot-id' in the PKCS#11 URI.
It is about bypassing searching the slot according to the token specified in the URI, if the slot is known in advance.

What do you think? I'm happy to receive comments.

I admit not having worked on tests yet. That I guess will have to follow... 

Best regards!

### DIFF
--- a/src/eng_back.c
+++ b/src/eng_back.c
@@ -365,6 +365,22 @@ int ctx_finish(ENGINE_CTX *ctx)
 }
 
 /******************************************************************************/
+/* Common functions for object handling                                       */
+/******************************************************************************/
+
+static int match_token(const PKCS11_SLOT* slot, const PKCS11_TOKEN* match_tok) {
+	return (match_tok && slot->token &&
+				(!match_tok->label ||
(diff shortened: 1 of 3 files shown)


mtrojnar:
Implementing `slot-id` is a good idea!

bordenc:
I've spent a few hours trying to figure this out (as I know little about PKCS 11, 15, etc.) I have an OpenPGP card where the key I need is on slot 1, and I can't seem to access it through OpenSSL using the URI syntax or `-key slot_1-id_01` syntax. This is because there's no way in this module to specify a slot?

mtrojnar:
@YetiFromTheAlps Are you still working on this draft PR?

PR #386, opened 2021-01-25
Fix issue with endianness when using addattr_bool

CallumMcLoughlin:
I have recently had problems when using this library on big endian systems (mips64, powerpc, etc) and have found the issue to be when using int as a boolean. The function sizeof(CK_BBOOL) returns 1, whereas int is not 1 byte (at least on my system).

This patch fixes an issue where adding a boolean attribute truncates the value and depending on the endianness of the system, results in the wrong value being added.

Alternatively, this fix could be implemented by changing the function signature?

I have tested on a number of little endian and big endian devices.

Thanks!


mtrojnar:
> Alternatively, this fix could be implemented by changing the function signature?

The fix you proposed is way more elegant.  Nice find!

PR #388, opened 2021-02-05
CLAIM_MODULE_TOKEN_FIPS

dengert:
In response to:
Re: [Opensc-devel]mutual authentication fails when FIPS enabled
and
https://bugzilla.redhat.com/show_bug.cgi?id=1827535

When using the OpenSSL FIPS library, with libp11 or engine,
a user can call PKCS11_set_claim_module_token_fips() to cause RSA_FLAG_FIPS_METHOD to be set.
Or when using the engine option CLAIM_MODULE_TOKEN_FIPS will also set RSA_FLAG_FIPS_METHOD.

By doing so the user/application is claiming the PKCS11 modules and tokens
to be used meet FIPS compliance.

It then becomes the responsibility of the user/application and not libp11.

Tested with OpenSSL-1.0.2 and OpenSSL-1.1.1 BUT NOT THE FIPS VERSION.

So consider this experimental and needs further testing.

For example:

openssl engine dynamic -vvvv -pre SO_PATH:$OPENSC_ENGINE/engines/engine_pkcs11.so -pre ID:pkcs11 -pre NO_VCHECK:1 -pre LIST_ADD:1 -pre LOAD -pre MODULE_PATH:$OPENSC_PATH/opensc-pkcs11.so -pre VERBOSE -pre FORCE_LOGIN -pre CLAIM_MODULE_TOKEN_FIPS

It can also be added to openssl.cnf like other pkcs11 engine options.

Please enter the commit message for your changes. Lines starting
 On branch CLAIM_MODULE_TOKEN_FIPS
 Changes to be committed:
	modified:   src/eng_back.c
	modified:   src/eng_front.c
	modified:   src/engine.h
	modified:   src/libp11.h
	modified:   src/p11_rsa.c


GeorgeC71:
This fixes my problem, which is described in
Re: [Opensc-devel]mutual authentication fails when FIPS enabled
and
https://bugzilla.redhat.com/show_bug.cgi?id=1827535

I did not observe any problems or errors associated with the fix. Everything looks good. 

I have tested the fix using OpenSSL 1.0.2u + OpenSSL FIPS  2.0.16.

dengert:
@Jakuje @ansasaki, See @GeorgeC71  comment https://github.com/OpenSC/libp11/pull/388#issuecomment-774552521

Jakuje:
Thank you for the PR. Let me check if it will work with our OpenSSL 1.1.1 on Monday.

src/eng_front.c
```diff
@@ -75,6 +75,10 @@ static const ENGINE_CMD_DEFN engine_cmd_defns[] = {
 		"FORCE_LOGIN",
 		"Force login to the PKCS#11 module",
 		ENGINE_CMD_FLAG_NO_INPUT},
+	{CMD_CLAIM_MODULE_TOKEN_FIPS,
+		"CLAIM_MODULE_TOKEN_FIPS",
+		"Take responsibility that Module and Token are FIPS complient",
```
ansasaki, line 80: Typo: s/complient/compliant/

ansasaki (commented):
I only found a small typo. Other than that, it LGTM

src/libp11.h
```diff
@@ -378,6 +378,7 @@ extern int PKCS11_generate_random(PKCS11_SLOT *slot, unsigned char *r, unsigned
 /*
  * PKCS#11 implementation for OpenSSL methods
  */
+extern void PKCS11_set_claim_module_token_fips(void);
```
mtrojnar, line 381: The `PKCS11_set_claim_module_token_fips()` function should also allow for turning this feature off (preferably, with a parameter).

mtrojnar, line 381: All exported symbols should be listed in `libp11.exports`.  In only works with our engine, because the engine is no longer linked dynamically with libp11.

src/eng_front.c
```diff
@@ -75,6 +75,10 @@ static const ENGINE_CMD_DEFN engine_cmd_defns[] = {
...
+		ENGINE_CMD_FLAG_NO_INPUT},
```
mtrojnar, line 81: Do we need an engine control command to turn this feature off (either with `ENGINE_CMD_FLAG_NUMERIC` or with a separate control command)?

mtrojnar (changes requested):
Nice work.  I like the idea of controlling the behavior with an engine control command.

dengert:
The problem with turning it off is there is only one RSA_METHOD created and used by all the RSA keys in libp11/engine. 

```
507 RSA_METHOD *PKCS11_get_rsa_method(void)
508 {
509         static RSA_METHOD *ops = NULL;
510 
511         if (!ops) {
```
And it is never freed either. 

It a perfect world, it should be possible to query PKCS11 if the module is FIPS compliant and is the token also compliant, but this is not possible today. So the choice was to keep the code simple and put the burden on the admin to "claim" the system is FIPS compliant (or in the spirit of FIPS). This was better then the RedHat mod to always set the flag. 
 
The current FIPS certified OpenSSL is based on 1.0.2. Not clear when the next one will be released, and will it force the use of EVP for RSA, or will it address the use of engines on hardware.

Jakuje:
If I read this right, this supports only modification of the mode only through the libp11 API calls or through openssl command. This makes it harder to use in generic applications using libp11 as an engine through openssl, which might not have an option to change this on their API, so the initial suggestion to have an option to configure this centrally inside of openssl or libp11 through some configuration sounded more appealing to me (rather than configuring all the applications separately).

dengert:
You can add the CLAIM_MODULE_TOKEN_FIPS  to openssl.cnf like any other engine command, such as the MODULE_PATH or FORCE_LOGIN as listed in   https://github.com/OpenSC/libp11#engine-controls

It is only needed when libp11 or engine is using the OpenSSL FIPS versions.  If you make this the default as the RedHat modification did, in my view that circumvents the deliberate requirement OpenSSL imposes on an RSA engine to require the RSA_FLAG_FIPS_METHOD be set, presumably to force the application developer/admin to say the engine meets the FIPS standards or spirit of FIPS. A distro should not be overriding this when it can not say the pkcs11 module or the token (or software implementation of a token) does not meet the FIPS standards. 
 
As a distro you could add this as a comment in the distributed openssl.cnf with the FIPS version. 
I see at least these 2 distributions (that was as far as I searched) have gotten FIPS certification:
https://www.redhat.com/en/about/press-releases/red-hat-completes-fips-140-2-re-certification-red-hat-enterprise-linux-7
https://ubuntu.com/blog/fips-certification-ubuntu-18-04-lts
and it looks like with OpenSSL-1.1.1. 

I don't think making RSA_FLAG_FIPS_METHOD a default would pass certification.  And libp11 should not set it as the RSA_FLAG_FIPS_METHOD even if a distro does.

src/eng_front.c
dengert: fixed.

src/libp11.h
dengert: Fixed.

Jakuje:
> You can add the CLAIM_MODULE_TOKEN_FIPS to openssl.cnf like any other engine command, such as the MODULE_PATH or FORCE_LOGIN as listed in https://github.com/OpenSC/libp11#engine-controls

Thank you for the pointer. I never used openssl engines so it is quite new for me.

> It is only needed when libp11 or engine is using the OpenSSL FIPS versions. If you make this the default as the RedHat modification did, in my view that circumvents the deliberate requirement OpenSSL imposes on an RSA engine to require the RSA_FLAG_FIPS_METHOD be set, presumably to force the application developer/admin to say the engine meets the FIPS standards or spirit of FIPS. A distro should not be overriding this when it can not say the pkcs11 module or the token (or software implementation of a token) does not meet the FIPS standards.
> 
> As a distro you could add this as a comment in the distributed openssl.cnf with the FIPS version.
> I see at least these 2 distributions (that was as far as I searched) have gotten FIPS certification:
> https://www.redhat.com/en/about/press-releases/red-hat-completes-fips-140-2-re-certification-red-hat-enterprise-linux-7
> https://ubuntu.com/blog/fips-certification-ubuntu-18-04-lts
> and it looks like with OpenSSL-1.1.1.
> 
> I don't think making RSA_FLAG_FIPS_METHOD a default would pass certification. And libp11 should not set it as the RSA_FLAG_FIPS_METHOD even if a distro does.

The way how FIPS certification works is that only fips modules are certified, where OpenSSL is one of them. The libp11 is not a FIPS module and therefore it does not undergo certification (since it does not implement any crypto on its own). The other FIPS module is the hardware/smart card and the wiring between them is at this moment missing and control from FIPS. But if I am right, that is going to be changed in FIPS 140-3, where we will need to provide some explicit signalization from the PKCS#11 modules for its users.

FYI, I tried to test this with our openssl on a system in FIPS mode with the libp11 testsuite and it looks like working fine also with 1.1.1. I will probably need to catch some more corner cases, but generally adding this to the openssl.cnf makes it working as expected.

dengert:
I agree with: "The other FIPS module is the hardware/smart card and the wiring between them is at this moment missing and control from FIPS. But if I am right, that is going to be changed in FIPS 140-3, where we will need to provide some explicit signalization from the PKCS#11 modules for its users."

The argument for not making RSA_FLAG_FIPS_METHOD the default, as it allows a misconception that  since OpenSSL is FIPS, anything it does is also FIPS.  If FIPS-140-3 is as you said, then they have recognized the problem too. The CLAIM_MODULE_TOKEN_FIPS is a stop gap measure, to get sites to make the evaluation of what they are doing is in the spirit of FIPS. 

https://csrc.nist.gov/projects/cryptographic-module-validation-program

https://csrc.nist.gov/projects/cryptographic-module-validation-program/validated-modules/search/all
shows 5 PIV applets on selected cards, 2 IBM PKCS\#11 modules, and many Red Hat and Ubuntu certiticates

But if: "we will need to provide some explicit signalization from the PKCS#11 modules for its users." This might imply OpenSC PKCS\#11 or a vendor's PKCS\#11 would also need to be certified.

mtrojnar:
Please resolve conflicts and describe the new engine control command in README.md.

mtrojnar:
The requested changes have not been implemented for almost a year.  As nobody seems to be interested in this feature, I'm closing the PR.

PR #390, opened 2021-03-03
load_key: Improve key loading by label. After importing a certificate…

justremotephone:
… on a SafeNet usb token under Windows the certificate is labeled but the key label is empty.

To solve this problem, the certificate label is used instead of empty key labels to find the key. Verbose output is also updated to show the effective key label and the matching certificate serial number is also written to the trace. So the key can be identified from the log output.


src/p11_cert.c
```diff
@@ -166,7 +166,7 @@ static int pkcs11_next_cert(PKCS11_CTX *ctx, PKCS11_TOKEN *token,
 	CK_OBJECT_HANDLE obj;
 	CK_ULONG count;
 	int rv;
-
+
```
mtrojnar, line 169: ?

PR #392, opened 2021-03-18
Ensure C_FindObjectsFinal is called in find_certs

eaaltonen:
If `C_FindObjectsInit` returns a non zero value, the previous version
would exit the function without calling `C_FindObjectsFinal`. Jump to
finalization step instead.

Signed-off-by: Eero Aaltonen <eero.aaltonen@vaisala.com>

### DIFF
--- a/src/p11_cert.c
+++ b/src/p11_cert.c
@@ -149,12 +149,17 @@ static int pkcs11_find_certs(PKCS11_TOKEN *token)
 	/* Tell the PKCS11 lib to enumerate all matching objects */
 	cert_search_class = CKO_CERTIFICATE;
 	rv = CRYPTOKI_call(ctx, C_FindObjectsInit(spriv->session, cert_search_attrs, 1));
-	CRYPTOKI_checkerr(CKR_F_PKCS11_FIND_CERTS, rv);
+	if (rv) {
+		CKRerr(CKR_F_PKCS11_FIND_CERTS, rv);
+		goto out;
+	}
+	ERR_clear_error();
 
 	do {
 		res = pkcs11_next_cert(ctx, token, spriv->session);
 	} while (res == 0);
 
+out:
 	CRYPTOKI_call(ctx, C_FindObjectsFinal(spriv->session));
 
 	return (res < 0) ? -1 : 0;


fabled:
My understanding is that if `C_FindObjectsInit` returns an error, the search operation was not initialized and thus there is no need to call the `Final` function.

dengert:
I agree. pkcs11 v2.40 says:
"Even though C_FindObjectsInit can return the values CKR_ATTRIBUTE_TYPE_INVALID and CKR_ATTRIBUTE_VALUE_INVALID, it is not required to. For example, if it is given a search template with nonexistent attributes in it, it can return CKR_ATTRIBUTE_TYPE_INVALID, or it can initialize a search operation which will match no objects and return CKR_OK."

But C_FindObjectsFinal can return CKR_OPERATION_NOT_INITIALIZED. i.e. C_FindObjectsInit must have returned an error, which should not cause the application to fail.

The real problem may be in `pkcs11_next_cert`
```
CK_OBJECT_HANDLE obj;
CK_ULONG count; 
```
Both  are not initialized, and may cause problems if some pkcs11 module do not clear them on error: 
```
rv = CRYPTOKI_call(ctx, C_FindObjects(session, &obj, 1, &count));
           CRYPTOKI_checkerr(CKR_F_PKCS11_NEXT_CERT, rv);
if (count == 0)
```

mtrojnar:
@dengert wrote:
> I agree.

Could you clarify what you agree with and what you recommend?

dengert:
Glad you asked. I Agree with @fabled, that if `C_FindObjectsInit` fails, there is no need to call `C_FindObjectsFinal`. Pkcs\#11 2.40 implies it is OK to call `C_FindObjectsFinal` as it can return `CKR_OPERATION_NOT_INITIALIZED`. But in this case it should not be needed. 

@eaaltonen 

Can you explain why you need to call `C_FindObjectsFinal` after `C_FindObjectsInit` fails?

What PKCS11 module are you using? 

Are you writing your own module?

Do you have a trace or log that shows why this PR is needed?  

Are you trying to call `pkcs11_find_certs`? 

(Disregard my comment about `pkcs11_next_cert`, I was not reading CRYPTOKI_checkerr correctly, which returns -1  on an error.)

dengert:
You should submit this as an issue to https://github.com/OP-TEE/optee_client. The misuse  of CKR_CANCEL for sure. They may not be handling C_Login when application has an open RO session. This would be promoted to a RW session by C_Login. The SPY log would show if this could be the problem.

Issue #343, opened 2020-07-01
OpenSSL fails to decrypt using EC keys...


llogar:
I've just noticed that on Ubuntu 20.04 OpenSSL decryption using EC keys fails with a _buffer to small_ error 
```
$ openssl cms -engine pkcs11 -decrypt -keyform engine -inkey "pkcs11:type=private;object=test" -inform PEM -in test.enc -recip test.crt
engine "pkcs11" set.
Enter PKCS#11 token PIN for SC-HSM (UserPIN):******
Error decrypting CMS using private key
139949706868032:error:0609909B:digital envelope routines:EVP_PKEY_derive:buffer too small:../crypto/evp/pmeth_fn.c:295:
```
This used to work just fine on Ubuntu 18.04 (with its version of libp11). I've traced the problem to EC pmeth announcing `EVP_PKEY_FLAG_AUTOARGLEN` (introduced in 0a2df89ba517bfbeaeadb81e42fe7bc3288b1985). Indeed, by removing this flag decryption works again.

As I am not sure, this is the correct way to fix this bug, I didn't make a pull request for the fix....

PR #394, opened 2021-03-20
Simplify pkey locking

fabled:
First set of clean ups to the code base in preparation for larger refactoring to allow proper multithreading support.

This simplifies the pkey internals to remove lock held over API calls. See each individual commit message for the detailed reasoning.


mtrojnar:
EVP_PKEY_FLAG_AUTOARGLEN may sound like a good idea, but sometimes it doesn't work as expected.  See: https://github.com/OpenSC/libp11/issues/343

fabled:
> EVP_PKEY_FLAG_AUTOARGLEN may sound like a good idea, but sometimes it doesn't work as expected. See: #343

This is enabled for RSA, but not for EC. This is correct and OpenSSL core RSA and EC methods do it exactly like this. This the changes to RSA method can assume it being enbled. If you see the last commit changing the `pkcs11_try_pkey_ec_sign` it actually fixes the function differently to support the size inquiry properly using the same pattern what OpenSSL core uses. Do note that this actually fixes the EC case, because it had incorrect check breaking the special case.

fabled:
>  See: #343

I can answer the question from #343 "As I am not sure, this is the correct way to fix this bug, I didn't make a pull request for the fix...."

The reason why this broke is that the P11 PKEY methods are "inherited" from the core OpenSSL PKEY methods by overriding only few functions. It calls on many paths the original OpenSSL functions. The OpenSSL EC PKEY method does not enable `EVP_PKEY_FLAG_AUTOARGLEN` because the way this auto calculation is implemented calculates wrong size for EC. If we just derive our method, and set this flag, the OpenSSL core ends up calling the original EC methods in incorrect way.

The proper fix is to clone the OpenSSL EC method's size checking code blob to our method which is done in the last commit. This fixes the function to work properly, and be able to give answers to the size inquiry without resorting to call PKCS11 library.

fabled:
Merged as part of #396.

Issue #393, opened 2021-03-19
thread locking issue crashes bind9/named


hcoin:
Multiple threads accessing private keys and adding/deleting pins needs more locking to prevent the crashes and core dumps detailed here:
https://bugzilla.redhat.com/show_bug.cgi?id=1940659

fabled:
There is various issues in this regard, and I am currently working on a fairly large refactoring PR to make libp11 more thread friendly. Auditing this concurrency issue is part of work. I hope to have a PR in few weeks time. The bug can be assigned to me by a admin if wanted.

hcoin:
I'm heavily rewriting eng_back.c, hopefully I'll have a version for you to look at that at least allows named to run without crashing in the context of several threads worth of dnssec updates/ongoing activity.

fabled:
I have now completed first good rounds of the refactoring: see https://github.com/OpenSC/libp11/pull/396

It adds various fixes to the thread safety, and also puts in rudimentary locking that should probably fix the issue you are having.

hcoin:
Take a look at these changes to eng_back.c  It consolidates much repeated code, adds two layers of locking, fixes a LibreSSL logic bug, passes all the tests, and -- named does not die under conditions of heavy load.

[eng_back.c.gz](https://github.com/OpenSC/libp11/files/6194122/eng_back.c.gz)

fabled:
There is definitely some good stuff in it. But as you also note, there's several things in one change: repeated code refactoring, locking changes, logic changes and the PIN allocation change. It would be helpful to have them as individual commits to help review.

I did modify the locking quite drastically in the referenced PR. However, you do additional locking in places I have not yet looked. So this probably fixes additional issues. I am also contemplating which is faster: enumerate all keys + O(n) lookup from the enumeration results; or just doing a PKCS#11 Find Object call with proper attributes for each URI. Considering the enumerate all keys currently refreshes the cache on every call, I think it would be faster to just do proper Find Objects with given attributes and drop the cache which is not really used. Or at least make it partial cache which contains only the items specifically searched for (not the full key data from the HSM which could be hundreds of keys in certain cases).

Please look at also the work I did. Perhaps doing this on top of the that branch? But at least I would hope to see PR with this split into 3-4 individual commits.

hcoin:
[...]

PR #396, opened 2021-03-22
Implement session pooling, add thread safety fixes and optimizations

fabled:
Improve performance and improve safety in multi threaded usafe of OpenSSL.
See individual commit messages for more detailed explanation of each change.

E.g. BIND9 dnssec-signzone test before these changes (on my i7-8 laptop):
```
Signatures generated:                    10216
Signing time in seconds:                13.666
Signatures per second:                 747.522
Runtime in seconds:                     13.726

real	0m 13.73s
user	0m 14.24s
sys	0m 1.25s

```

And after:
```
Signatures generated:                    10216
Signing time in seconds:                 4.779
Signatures per second:                2137.268
Runtime in seconds:                      4.833

real	0m 4.84s
user	0m 21.88s
sys	0m 2.53s
```


mtrojnar:
The performance of you PKCS#11 token is really impressive.  Which device do you use?
Have you verified that the produced signatures are correct?  Have you tried any other devices?

fabled:
> The performance of you PKCS#11 token is really impressive. Which device do you use?

Results are from softhsm2 (openssl build) with RSA 2K keys.

> Have you verified that the produced signatures are correct? Have you tried any other devices?

I did thorough analysis of the libp11 and relevant openssl code. I have quite a bit of experience implementing PKCS#11 tokens, and using them. So special attention has been paid to correctness. And as documented, this also fixes several correctness issues. There's still some potential fixes pending, but will address them in future PRs.

This PR passes test suite as correctness proof. And running `dnssec-signzone` with `-a` to verify the signature will confirm all signatures are good:
```
Signatures successfully verified:        10216
```

Other devices are not yet tested, but is planned against a high-end external HSM device. Can report back when results are available.

dengert:
This PR should also be tested against low tech devices and other PKCS\#11 modules, that may not have thread protection, depending on libp11 single threading calls to these modules.

This may not be a problem, as I don't see any changes to pkcs11_CTX_load where C_Initialize is called and return values are checked. But based on other discussions, the intent of this change it to queue requests for crypto operations for the same key as close to the hardware as possible, which could be in a card driver or the card itself, possible by using multiple channels to the card. 

https://github.com/johughes99  on OpenSC lists you have said you have been testing PKCS\#11 modules. Do you have any comments on the PR?

fabled:
Yes, further testing is definitely needed.

Regarding the assumptions made, the PKCS#11 v2.40 states the following:
 - Base Spec 5.4 C_Initialize:
    >A call to C_Initialize specifies one of four different ways to support multi-threaded access via the value of the CKF_OS_LOCKING_OK flag in the flags field and the values of the CreateMutex, DestroyMutex, LockMutex, and UnlockMutex function pointer fields:
    >...
    > 2.     If the flag is set, and the function pointer fields aren’t supplied (i.e., they all have the value NULL_PTR), that means that the application will be performing multi-threaded Cryptoki access, and the library needs to use the native operating system primitives to ensure safe multi-threaded access.  If the library is unable to do this, C_Initialize should return with the value CKR_CANT_LOCK.
    And We pass this flag, so the library should return error if it does not support errors.
 - Usage Guide 2.5.3
   > In particular, when an application initializes a Cryptoki library with a call to C_Initialize, it can specify one of four possible multi-threading behaviors for the library:
   > ...
   > 2.       The application can specify that it will be accessing the library concurrently from multiple threads, and the library must be able to use native operation system synchronization primitives to ensure proper thread-safe behavior.
 - Usage Guide 4.1
   > Different threads of an application should never share sessions, unless they are extremely careful not to make function calls at the same time.  This is true even if the Cryptoki library was initialized with locking enabled for thread-safety.

Of course older libraries might comply to older versions of this. But my current understanding is that, also with the older spec versions `CKF_OS_LOCKING_OK` means the library must go into multi-threaded mode or return error.

And as noted in the Usage Guide, the rule is that multi threaded access must never use same session in multiple threads concurrently. Session is the per-thread context concept. Same session can be used in multiple threads though when application synchronizes things so that it's not used concurrently. This is the approach selected in this PR: a session pool is maintained and sessions are checked out when needed, and returned after use.

The Usage Guide is also very clear that same object handles can are shareable between sessions.

This was the simplest way to implement proper multithreaded support. And it conforms to the PKCS#11 standard. If some PKCS#11 module breaks, it is likely non-conforming and should be fixed. We can of course add additional knobs to adjust the session pool size so that a single threaded access is forced globally if needed.

But obviously this PR is a non-trivial, complex and in some parts intrusive change set. So thorough review and testing is needed. I am sure there are points were changes will be required. I tried to put most thing in isolated individual commits so it might make sense to review and merge this in sets. There are also few fairly large commits that refactor some specific API.

And as explained in the commits, this series even fixes some issues where the single session was "accidentally" used concurrently from multiple threads. Causing crashes such as #393.

I have low-end HSM device here, soon two different ones. And will be testing them soon. But wanted to share this early for review. I am happy to fix any issues found.

dengert:
Thanks, for the quick response. 

The way I read PKCS\#11 v2.20, is if CK_C_INITIALIZE_ARGS_PTR
is NULL, "the application will not be accessing Cryptoki through multiple threads simultaneously." If it is not NULL, then the application is asking if the module can support
either CKF_OS_LOCKING_OK or use the routines provided by the application.  
If the module can't do one of these it should return CKR_CANT_LOCK
(Or return CKR_NEED_TO_CREATE_THREADS if the module needs to use threads and the application told it to not use threads via CKF_LIBRARY_CANT_CREATE_OS_THREADS). 

So that should not be a problem, as current libp11 always passes in CKF_OS_LOCKING_OK, but  I have not looked closely how libp11 actually uses threads. 

Looks like you are addressing this concern.

mtrojnar:
We are testing it with stunnel. I should have some news soon.

fabled:
AEP Keyper HSM PKCS#11 (with load-balancer support) was tested now. No bugs or issues found. Performance differences as follows (with similar dnssec-signzone test):
1. RSA keys: before patch ~250 signatures/s, with PR 1333 signatures/s
2. EC keys: before PR ~210 signatures/s, with PR applied 615 signatures/s

Showing 3-5x speedup.

Testing did reveal that the operations are not distributed evenly to the load balancing cluster. So I will write a follow up commit to change the current LIFO strategy to FIFO in the session pooling.

Also comparison against other software implies there is still potential for increased performance. I will do some additional analysis if something can be still improved in the hot crypto paths, and followup with separate commits as needed.

mtrojnar:
The initial testing results are quite positive.  We could not observe any speedup with stunnel, but I guess the usage pattern in TLS might be different from dnssec-signzone.

The one change I don't like is the replacement of the portable OpenSSL synchronization API with pthreads.  The rationale for using OpenSSL was that no additional changes are needed to libp11 after OpenSSL and the PKCS#11 module are ported to the new platform (possibly, an obscure embedded OS). Also, I don't really like the idea of emulating pthreads API with the Windows API.  Is there a significant performance gain of using pthreads directly?  Would you consider reverting this part of your PR?

Otherwise, I really like the code and I'll be glad to merge your PR.

fabled:
The rationale for doing the pthread dependency / emulation is that I need the "condition variable" primitive for session pool synchronization. This is not available via the OpenSSL primitives. This could be emulated with busy-polling but would likely not have equal performance gain.

Do note that current OpenSSL unconditionally uses the same pthreads and Windows APIs, so having OpenSSL available, basically means the threading dependency is satisfied. The third it supports is 'no threading' where the locking is stubs doing nothing.

I would really like to keep the pthread condition variable approach. And if pthreads/Windows threads is not available, we could fall back to building non-threaded library (pthread stubs that do nothing) similar to OpenSSL.

I am open to reworking the code. But would then need suggestions how to implement condition variables in acceptable manner. And if you'd then instead prefer some internal wrappers that use pthreads or Wndows threads? I found doing the pthreads wrappers for Windows simplest choice. But agreeably it can confuse some.

mtrojnar:
Good point.  I don't know a good way to implement a resource pool without condition variables.

mtrojnar:
> The rationale for doing the pthread dependency / emulation is that I need the "condition variable" primitive for session pool synchronization. This is not available via the OpenSSL primitives.

@fabled How should we deal with languages that provide their own lightweight multiprocessing, such as Erlang or Go?
See https://github.com/OpenSC/libp11/issues/407 for more details.

fabled:
> @fabled How should we deal with languages that provide their own lightweight multiprocessing, such as Erlang or Go?
> See #407 for more details.

Depends on the threading model. Erlang seems to be green threading (1 OS thread to N Erlang threads). In this case using the pthreads is ok, also bypassing locking would be ok too. Of course session pooling will not help much here, because all operations will be multiplexed to the same OS thread.

I think pthreads should work for Go too. While Go support M:N threading, the calls to libp11 do all the locking internally. That is Go will not do OS thread switch for libp11 stacks. In this scenario pthreads locking should suffice.

ebenwar:
FYI, on a multi-core system Erlang by default has one scheduler(OS-thread) per cpu.

PR #398, opened 2021-03-30
add re-numeration of slots as engine ctrl command

commonism:
This was broken in  14cd0d328fff96b79fabcc30257e358399c8ad25.
Previously, the engine would re-enumerate before loading keys/certs
Not re-enumerating the slots results in un-awareness of changes in slots
and tokens.
This awareness is required to be able to change the token in a slot at
runtime, else you use invalid sessions
(PKCS#11 module:pkcs11_find_keys:Session handle invalid:p11_key.c:512)

The patch adds the command RE_ENUMERATE as engine control, providing the
ability to re-enumerate on demand/when required.


mtrojnar:
Please also describe the new engine control command in README.md.

Issue #407, opened 2021-05-21
One process multiple workloads issue


Eliot-Roxbergh:
I have one process which may have multiple workloads at the same time interacting with libp11 (e.g. Erlang VM running multiple workloads, which to the OS looks like one process).
However, I run into problems with both libp11 C API and libp11_engine (two unrelated issues).

**Without engine;**
The single process is calling libp11, however, it has multiple workloads active at the same time - and needs to call PKCS11_enumerate_slots for each.
So there are multiple PKCS11_CTX present for the single process, each needs to be freed at some point and their sessions released by calling PKCS11_release_all_slots.
This creates a problem, when the first workload does this and exits, the other's sessions are invalidated. (at the same time, release_all_slots needs to be called for each workload to free their memory.)
_"C_CloseAllSessions closes all sessions an application has with a token"_

As far as I can see, this would be resolved if pkcs11_release_slot only issues C_CloseSession for its own active sessions (as set in PKCS11_CTX), and not as currently done with C_CloseAllSessions.

**With engine;**
When using the engine, I instead get another problem because it only reads the slot information once.
For instance, if a key is written to the HSM after the engine has already been is initialized, this key will not be found. Either the process needs to restart, or the application manually call RE_ENUMERATE.

1) start (long-lived) Erlang process
2) generate keys for workload A
3) start workload A
    p11-engine asks for all slots and gets A's key
4) generate keys for workload B
5) start workload B
6) error -> p11-engine, slots are already loaded but B's key cannot be found

dengert:
What versions of libp11,  engine, OpenSSL are you using?
What pkcs11 module(s) are you using? 

PKCS11 is designed for only one instance of the lib per application (process). So having "multiple PKCS11_CTX present for the single process" may not appropriate. pkcs11_release_slot applies to all sessions for that slot. 

Maybe your workloads need to share a single PKCS11_CTX per slot and  pkcs11_release_slot only gets called when the application is done with the slot. Note the login state is per slot also. 

Most PKCS11 modules supports threads. Does the Erlang VM support threads? 

These may also related:
https://github.com/OpenSC/libp11/issues/365
https://github.com/OpenSC/libp11/pull/396

Eliot-Roxbergh:
libp11 (& engine) = 779350e92f47f1d262f6b7e4c5e86c7b1553918a
OpenSSL = 1.1.1g
SoftHSMv2 = 2.6.1+

[...]

PR #408, opened 2021-05-27
ctx->slot_list improvements

### LINKED ISSUE #403: Potential memory leak
I noticed the new feature CMD_RE_ENUMERATE, commit [add re-numeration of slots as engine ctrl command ](https://github.com/OpenSC/libp11/commit/bfcd96a58ad3535a68a3b599a11e9f9f2e568e59) and I think it may result in memory leaks.

PKCS11_enumerate_slots is called in

`static  int ctx_enumerate_slots(ENGINE_CTX *ctx, PKCS11_CTX *pkcs11_ctx)`

to read the current slot list and count from the hsm which is then assigned to the passed ENGINE_CTX ctx.
if `ctx->slots` has been assigned in a previous call memory is leaked.
Note: I guess `PKCS11_release_all_slots` can't be called at this point since the old list might be in use by another thread.

dengert:
This is only partial tested, and need testing by those who have this problem.

free ctx->slot_list when it is being replaced to avoid memory leak.
Fixes #403
Partial fix for #407

Add mutex calls for ctx->lock when needed.

Other routines that use the ctx->slot_list are
called with ctx->lock held. But ctx_enumerate_slots is called
from two places,ctx_init_libp11_unlocked  with the lock held
and ctx_engine_ctrl: CMD_RE_ENUMERATE which is not holding the lock.

 On branch slot_list
 Changes to be committed:
	modified:   eng_back.c

### DIFF
--- a/src/eng_back.c
+++ b/src/eng_back.c
@@ -248,16 +248,20 @@ int ctx_destroy(ENGINE_CTX *ctx)
 	return 1;
 }
 
-static int ctx_enumerate_slots(ENGINE_CTX *ctx, PKCS11_CTX *pkcs11_ctx)
+static int ctx_enumerate_slots(ENGINE_CTX *ctx, PKCS11_CTX *pkcs11_ctx, int need_lock)
 {
 	PKCS11_SLOT *slot_list = NULL;
 	unsigned int slot_count = 0;
 
 
+	if (need_lock)
+		pthread_mutex_lock(&ctx->lock);
 	/* PKCS11_CTX_load() uses C_GetSlotList() via p11-kit */
 	if (PKCS11_CTX_load(pkcs11_ctx, ctx->module) < 0) {
 		ctx_log(ctx, 0, "Unable to load module %s\n", ctx->module);
 		PKCS11_CTX_free(pkcs11_ctx);
+		if (need_lock)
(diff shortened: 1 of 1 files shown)


src/eng_back.c
```diff
@@ -1085,7 +1094,7 @@ int ctx_engine_ctrl(ENGINE_CTX *ctx, int cmd, long i, void *p, void (*f)())
 	case CMD_FORCE_LOGIN:
 		return ctx_ctrl_force_login(ctx);
 	case CMD_RE_ENUMERATE:
-		return ctx_enumerate_slots(ctx, ctx->pkcs11_ctx);
+		return ctx_enumerate_slots(ctx, ctx->pkcs11_ctx, 1);
```
fabled, line 1097: Why not just add locking around this call? This would make this single hunk change. Or add separate helper function that does slot enumeration with lock held. Adding flag whether or not to do locking is usually error prone.

dengert: That would work too. Go ahead and add equivalent code to your other changes. 

I am not sure if I agree with: "Adding flag whether or not to do locking is usually error prone." Not having the flag got to this error, that ctx_enumerate_slots could be called while some other  thread was using the `ctx->slot_list` to load keys or certificates.

PR #409, opened 2021-05-30
Simplify engine code to load certificates, public and private keys

fabled:
It was mostly copy paste, so remove the duplicate code to simplify
code maintenance and readability. The per-slot object matching code
is made a callback because the internally cached structs do not
share a common header portion. Make the helper functions do as much
as possible to minimize duplicate code.

This implements similar work done by @hcoin in #393, but takes slightly
different approach.


fabled:
@mtrojnar Ping. Have you had time to look at this yet? This would be basis for the follow up work related to #410.

mtrojnar:
I reviewed the code and I like it.  Using `void *` is not my preferred coding style, but alternatives don't seem much better.

PR #415, opened 2021-06-28
Make PKCS11_KEY_ops definitions consistent

fabled:
This simplifies the code a bit, as pkcs11_ec_ops is now defined
directly without the pointer indirection. The sole place of use
now uses OPENSSL_NO_EC instead.


fabled:
Part of #414 now.

PR #425, opened 2021-09-24
Expand the PKCS#11 engine configuration instructions

oerdnj:
On Debian, existing ``openssl_conf = default_conf`` would override the
``openssl_conf = openssl_init`` line needed to initialize engine.  This
leads to hard to debug state, where the ``[openssl_init]`` section gets
completely ignored.

Add more specific instructions that the ``openssl_conf`` line must be
defined only once because the later override the former.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,15 @@ before any sections are defined:
 openssl_conf = openssl_init
 ```
 
+WARNING: There must be only a single ``openssl_conf = `` line because the later
+will override former.  If there's an existing ``openssl_conf`` directive either
+edit it or commment out the previous definition:
+
+```
+# System default
+# openssl_conf = default_conf
+```
+
 This should be added to the bottom of the file:
 
 ```


dwmw2:
Ick. Aren't all those instructions just wrong anyway?

If you need *anything* other than just using a PKCS#11 URI in place of a filename, then your distribution is Getting It Wrong™.

If it *needs* documenting, fix it and document what's left.
Complex documentation should be the last resort for the things we *can't* fix.

oerdnj:
> If you need anything other than just using a PKCS#11 URI in place of a filename, then your distribution is Getting It Wrong™.

I don't follow - how exactly am I supposed to pass the integration with the softhsm2 and pkcs11 engine compiled from source in non-standard location into PKCS#11 URI?

dwmw2:
Installing providers is a matter for the p11-kit configuration; the engine just defaults to using `p11-kit-proxy.so` as its provider so that should all work out.

The PKCS#11 engine can be installed into the system directory (e.g. `/usr/lib/engines-1.1`) and should work fairly much automatically, shouldn't it?

oerdnj:
> Installing providers is a matter for the p11-kit configuration; the engine just defaults to using `p11-kit-proxy.so` as its provider so that should all work out.

No, I don't think this is a "one size fits all" situation. The module provides a way to use PKCS#11 modules directly, requiring `p11-kit-proxy.so` would add an extra layer with no real reason.

> The PKCS#11 engine can be installed into the system directory (e.g. `/usr/lib/engines-1.1`) and should work fairly much automatically, shouldn't it?

What you are saying is that any use of _engine_pkcs11_ requires to have root access. That's not true and unnecessary.

This MR documents a common pitfall and you are arguing to just makes this more obscure. I really don't follow.

dwmw2:
> What you are saying is that any use of engine_pkcs11 requires to have root access. That's not true and unnecessary.

I'm a bit confused. Isn't the part you're *modifying* the bit where you have to have root access to edit the global OpenSSL config?

I'm all for clearly documenting how non-root users can install the engine for themselves on a system which doesn't have it — but is that what you've done?

oerdnj:
> I'm a bit confused. Isn't the part you're _modifying_ the bit where you have to have root access to edit the global OpenSSL config?

No, setting `OPENSSL_CONF` to a custom location works just fine.

> I'm all for clearly documenting how users can install the engine for themselves on a system which doesn't have it, and one which they don't have root access — but is that what you've done?

That's not a goal of this MR. The goal of this MR is prevent people to burn several hours before they notice the extra configuration isn't doing anything because there's extra `openssl_conf =` line in the configuration.

dwmw2:
I think that's best achieved by deleting the whole section. It's leading people in the wrong direction, documenting on the front-page README.md how to use a non-recommended solution.

This all ought to have worked out of the box with the distribution's packaged versions of everything.

oerdnj:
> This all ought to have worked out of the box with the distribution's packaged versions of everything.

Does it?

> I think that's best achieved by deleting the whole section.

I would argue that deleting the whole section without a viable working replacement isn't really an improvement. While this MR provides the immediate solution for a problem that more people could encounter out of the box.

Nevertheless, this consumed more time than I want to spend on a single paragraph of text, so closing the MR.

dwmw2:
> > This all ought to have worked out of the box with the distribution's packaged versions of everything.

> Does it?

It certainly does in Fedora, and the packaging guidelines there strongly encourage packagers to ensure that things *do* Just Work when users provide a PKCS#11 URI in place of a filename for certificates/keys/etc.

If it doesn't in Debian, the best place to file a bug is Debian's own bug tracker. And the upstream pkcs11_engine documentation should probably make the expectations clearer — in a wel-maintained distribution, this stuff *should* Just Work.

PR #433, opened 2021-12-07
Tighten up debug logging / Wrong certificate returned if multiple certs have same label but different ID

minfrin:
Make lists more obvious. Include details of the module being loaded,
the tokens being scanned, and the reasons for not finding objects.

Debug logging used to look like this (in v4.10)

```
PKCS#11: Initializing the engine
Found 3 slots
Loading private key "pkcs11:token=softtok;id=%B0%90%5A%36%1A%E0%2F%DA%13%14%81%FC%E1%DE%9D%86%9E%30%8F%AE;object=;type=private?pin-value=1111"
Loading private key "pkcs11:token=softtok;id=%B0%90%5A%36%1A%E0%2F%DA%13%14%81%FC%E1%DE%9D%86%9E%30%8F%AE;object=;type=private?pin-value=1111"
Looking in slot -1 for key: id=b0905a361ae02fda131481fce1de9d869e308fae label=
[17] Linux                      uninitialized, login  (softtok)
Found uninitialized token
[18] SoftHSM slot ID 0x6be9cd2  login             (SoftHSM Token)
[19] SoftHSM slot ID 0x1        uninitialized, login  (no label)
Specified object not found
PKCS11_get_private_key returned NULL
```

Now it looks like this:

```
PKCS#11: Initializing the engine: /usr/lib64/p11-kit-proxy.so
Found 3 slots
Looking in slots for private key without login: id=b0905a361ae02fda131481fce1de9d869e308fae label=
- [17] Linux                      uninitialized, login                  (softtok)
- [18] SoftHSM slot ID 0x6be9cd2  login                                 (SoftHSM Token)
- [19] SoftHSM slot ID 0x1        uninitialized, login                  (no label)
No matching initialized token was found for private key
Looking in slots for private key with login: id=b0905a361ae02fda131481fce1de9d869e308fae label=
- [17] Linux                      uninitialized, login                  (softtok)
- [18] SoftHSM slot ID 0x6be9cd2  login                                 (SoftHSM Token)
- [19] SoftHSM slot ID 0x1        uninitialized, login                  (no label)
No matching initialized token was found for private key
The private key was not found at: pkcs11:token=softtok;id=%B0%90%5A%36%1A%E0%2F%DA%13%14%81%FC%E1%DE%9D%86%9E%30%8F%AE;object=;type=private?pin-value=1111
PKCS11_get_private_key returned NULL
```

When ID and label are specified, both need to match, not either.
To fix this id-match OR label-match was replaced with id-match AND
label-match.

A tiebreak was added when multiple matching certificates could be
returned. The certificate with the latest expiry wins, and if we
still have a tie we deterministically choose a certificate using
X509_cmp().

If we do not specify a certificate, we return the first certificate
(or first certificate with an ID) as before.

Debug logging updated to show the expiry date used in the decision.

### DIFF
--- a/src/eng_back.c
+++ b/src/eng_back.c
@@ -298,7 +298,7 @@ static int ctx_init_libp11_unlocked(ENGINE_CTX *ctx)
 	if (ctx->pkcs11_ctx && ctx->slot_list)
 		return 0;
 
-	ctx_log(ctx, 1, "PKCS#11: Initializing the engine\n");
+	ctx_log(ctx, 1, "PKCS#11: Initializing the engine: %s\n", ctx->module);
 
 	pkcs11_ctx = PKCS11_CTX_new();
 	PKCS11_CTX_init_args(pkcs11_ctx, ctx->init_args);
@@ -388,6 +388,8 @@ static void *ctx_try_load_object(ENGINE_CTX *ctx,
 					ctx->pin_length = tmp_pin_len;
 				}
 			}
+			ctx_log(ctx, 1, "Looking in slots for %s %s login: ",
+				object_typestr, login ? "with" : "without");
 		} else {
 			n = parse_slot_id_string(ctx, object_uri, &slot_nr,
 				obj_id, &obj_id_len, &obj_label);
@@ -401,9 +403,9 @@ static void *ctx_try_load_object(ENGINE_CTX *ctx,
 				ENGerr(ENG_F_CTX_LOAD_OBJECT, ENG_R_INVALID_ID);
 				goto error;
 			}
+			ctx_log(ctx, 1, "Looking in slot %d for %s %s login: ",
+				slot_nr, object_typestr, login ? "with" : "without");
 		}
-		ctx_log(ctx, 1, "Looking in slot %d for %s: ",
-			slot_nr, object_typestr);
 		if (obj_id_len != 0) {
 			ctx_log(ctx, 1, "id=");
 			dump_hex(ctx, 1, obj_id, obj_id_len);
@@ -457,7 +459,7 @@ static void *ctx_try_load_object(ENGINE_CTX *ctx,
 					!strcmp(match_tok->model, slot->token->model))) {
 			found_slot = slot;
 		}
-		ctx_log(ctx, 1, "[%lu] %-25.25s  %-16s",
+		ctx_log(ctx, 1, "- [%lu] %-25.25s  %-36s",
 			PKCS11_get_slotid_from_slot(slot),
 			slot->description, flags);
 		if (slot->token) {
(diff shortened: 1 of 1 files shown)


src/eng_back.c
```diff
@@ -605,11 +644,45 @@ static void *ctx_load_object(ENGINE_CTX *ctx,
...
+		return a;
+	}
+
+	aa = X509_get0_notAfter(a->x509);
+	ba = X509_get0_notAfter(b->x509);
+
+	/* the best certificate expires last */
+	if (ASN1_TIME_diff(&pday, &psec, aa, ba)) {
+		if (pday < 0 || psec < 0) {
+			return a;
```
Jakuje, line 666: If you check for the end of expiration date and select the later one, you should check also for the start of validity. In rare case, you might get certificate with validity starting in the future and this cmp function will select it.

src/eng_back.c
```diff
@@ -618,28 +691,66 @@ static void *match_cert(ENGINE_CTX *ctx, PKCS11_TOKEN *tok,
...
+			} else if (obj_label && !obj_id_len) {
+				if (k->label && strcmp(k->label, obj_label) == 0) {
+					selected_cert = cert_cmp(selected_cert, k, NULL);
+				}
+			} else if (obj_id_len && !obj_label) {
+				if (k->id_len == obj_id_len &&
+						memcmp(k->id, obj_id, obj_id_len) == 0) {
+					selected_cert = cert_cmp(selected_cert, k, NULL);
+				}
+			}
```
Jakuje, line 721: I am just wondering if this could be written with less code duplication. Maybe just:
```

			if ((!obj_label || k->label || (k->label && strcmp(k->label, obj_label) == 0) &&
			   (!obj_id_len || k->label || (k->id_len == obj_id_len && memcmp(k->id, obj_id, obj_id_len) == 0)) {
 				selected_cert = cert_cmp(selected_cert, k, NULL);
 			}
```
Its have sligltly different semantics and not sure if it is more readable, but just for consideration ..

src/eng_back.c
minfrin: My original attempt to fix this took into account the start of validity, and then I realised to make a decision with respect to the start of validity you need the current time, but there was no concrete mechanism I could use to guarantee that I could get the correct time.

For low security stuff the system clock might be enough, for higher security stuff where the HSM has a clock on board getting access to that clock reliably was something I couldn't figure out. I didn't want any behaviour that wasn't 100% deterministic.

I eventually settled on expiry date only, as that comparison always yields the same winner. The cost of this is that people deploying certs shouldn't deploy certs that aren't valid yet, which seems to be a sensible limitation, and if you do need to distinguish in this case where two certs issued at different times have the same ID (potentially because same subject key identifier because same key), use a label.

src/eng_back.c
minfrin: I was going for readability, so it's clear which of three cases result in the cert being chosen as the current winner.

Jakuje (approved):

PR #434, opened 2021-12-07
Add missing files to autotools: "make dist" works again.

minfrin:
When "make dist" is used to generate a tarball, a header file and a test case is missing, causing build failure.

This patch adds the missing files to autotools, and the tarball generated by "make dist" works.


Jakuje:
It would be great if libp11 would some ci with github actions that could run the `make dist` with each PR. It should be quite easy also with running tets in different configurations.

mtrojnar:
@Jakuje Good point.  I'm working on it.

minfrin:
Quick ping on this one, any news?

Issue #439, opened 2022-03-11
OAEP: oaep->source should be set to 1UL not 0UL


space88man:
In the implementation of OAEP we have p11_pkey.c 

        /* we do not support the OAEP "label" parameter yet... */
        oaep->source = 0UL; /* empty parameter (label) */
        oaep->pSourceData = NULL;
        oaep->ulSourceDataLen = 0;

However the only accepted value in PKCS#11 is `oaep->source = CKZ_DATA_SPECIFIED // 1UL`; absence of label is conveyed by  subsequent `NULL` and `0`.

For a strict HSM (Thales Luna) using pkcs11 engine for OAEP decryptions fails due to `CKR_MECHANISM_INVALID`.  

    # openssl exemplar
    openssl pkeyutl -decrypt -keyform engine -engine pkcs11 ...
        -pkeyopt rsa_padding_mode:oaep
        -pkeyopt rsa_oaep_md:sha384 -pkeyopt rsa_mgf1_md:sha384

    # rejected CK_RSA_PKCS_OAEP_PARAMS
    "60020000000000000300000000000000000000000000000000000000000000000000000000000000"
    
However Thales vendor engine is sending:

    # accepted CK_RSA_PKCS_OAEP_PARAMS
    "60020000000000000300000000000000010000000000000000000000000000000000000000000000"

Notice the single byte difference: "01" vs "00"

Another working library, python_pkcs11 always sets

    oaep_params.source = CKZ_DATA_SPECIFIED # this value is 1UL

PR #440, opened 2022-03-11
Fix OAEP source param

space88man:
The only supported value for the source field of OAEP params is `1UL` or  `CKZ_DATA_SPECIFIED`.

Some HSMs(Thales Luna) are strict about enforcing this and reject `0UL` as an invalid mechanism.

It could be that softhsm2 is more lenient about parsing this parameter.

Addresses #439

Update: indeed softhsm2 does not check this parameter for `CKZ_DATA_SPECIFIED` during decryption

### DIFF
--- a/src/p11_pkey.c
+++ b/src/p11_pkey.c
@@ -301,8 +301,8 @@ static int pkcs11_params_oaep(CK_RSA_PKCS_OAEP_PARAMS *oaep,
 	if (!oaep->hashAlg || !oaep->mgf)
 		return -1;
 	/* we do not support the OAEP "label" parameter yet... */
-	oaep->source = 0UL; /* empty parameter (label) */
-	oaep->pSourceData = NULL;
+	oaep->source = CKZ_DATA_SPECIFIED;
+	oaep->pSourceData = NULL; /* empty parameter (label) */
 	oaep->ulSourceDataLen = 0;
 	return 0;
 }


src/p11_pkey.c
```diff
@@ -300,9 +300,9 @@ static int pkcs11_params_oaep(CK_RSA_PKCS_OAEP_PARAMS *oaep,
 	oaep->mgf = pkcs11_md2ckg(mgf1_md);
 	if (!oaep->hashAlg || !oaep->mgf)
 		return -1;
+	oaep->source = 1UL; /* CKZ_DATA_SPECIFIED */
```
Jakuje, line 303: Please, use the constant directly -- its defined in the `pkcs11.h`. Using magic numbers leads only to errors like this.

https://github.com/OpenSC/libp11/blob/5e25cc823e6d569bcb35203b2b2c595086adf7e4/src/pkcs11.h#L869

Jakuje:
I am wondering why this is not caught by the tests that try the OAEP encryption. The softhsm should be checking for this since 2014.

Jakuje:
> I am wondering why this is not caught by the tests that try the OAEP encryption. The softhsm should be checking for this since 2014.

Sigh ... the check is there for encryption, wrapping, unwrapping, but not for decryption ... filled https://github.com/opendnssec/SoftHSMv2/pull/671

src/p11_pkey.c
space88man: Done — force pushed

space88man: Done — force pushed

PR #445, opened 2022-04-17
When a PIN explicitly provided use a PIN regadless of secure login flag

alonbl:
When a token supports both protected authentication path and PIN the user
may choose which method it wishes to use in order to authenticate.

The default is of course the best secure method, however, when a user
specified a PIN either explictly or via the URI then respect this setting.

Signed-off-by: Alon Bar-Lev <alon.barlev@gmail.com>

### DIFF
--- a/src/eng_back.c
+++ b/src/eng_back.c
@@ -49,6 +49,7 @@ struct st_engine_ctx {
 	 */
 	char *pin;
 	size_t pin_length;
+	int forced_pin;
 	int verbose;
 	char *module;
 	char *init_args;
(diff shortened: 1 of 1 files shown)


alonbl:
Codespell fix is at https://github.com/OpenSC/libp11/pull/446

src/eng_back.c
```diff
@@ -941,6 +942,7 @@ static int ctx_ctrl_set_pin(ENGINE_CTX *ctx, const char *pin)
 	 * shall be returned and errno shall be set. */
 	ctx_destroy_pin(ctx);
 	ctx->pin = OPENSSL_strdup(pin);
+	ctx->forced_pin = 1;
```
Jakuje, line 945: nit: I would probably move this line after the allocation check or the length count to make the code easier to read and keep the pieces that belong together together.

alonbl: Done

Issue #456, opened 2022-07-25
OpenSSL 3: Enabling pkcs11 engine makes EC operations fail


vuori:
I built the libp11 and openssl master branches last night, and it seems that the presence of the pkcs11 engine causes problems with various operations on EC keys. Note that the issue occurs if the engine is just present in the configuration, even if it's not being used.

Steps to reproduce:

  1. Copy the default `openssl.cnf` from OpenSSL sources, insert relevant `engine_section` and `pkcs11_section` bits from README with `dynamic_path` pointing to freshly-built `pkcs11.so`.
  2. Try to generate an EC key: `openssl ecparam -name brainpoolP256r1 -genkey -out t1.key`. Note that it fails with error `80A26D89627F0000:error:078C0102:common libcrypto routines:get_string_internal:passed a null parameter:crypto/params.c:1258:`. Comment out `pkcs11 = pkcs11_section`, try again and the key is generated.
  3. Re-enable the engine and try to generate a self-signed cert: `openssl req -subj /CN=ectest -new -x509 -key t1.key -out t1.crt`. Error `80B2C6CB5A7F0000:error:03000093:digital envelope routines:default_check:command not supported:crypto/evp/ctrl_params_translate.c:329:` occurs. Disabling the engine allows creating the cert.
  4. Commands that shouldn't ever need the engine such as `openssl cms -encrypt t1.crt` where `t1.crt` is an EC cert also fail with a similar error: `801234265C7F0000:error:03000093:digital envelope routines:evp_pkey_ctx_ctrl_int:command not supported:crypto/evp/pmeth_lib.c:1321:`

Though the messages are different, this may be related to issue https://github.com/OpenSC/libp11/issues/455. RSA operations work fine, including ones that use a smartcard through the pkcs11 engine.

ch-f:
This seems to be a duplicate of https://github.com/OpenSC/libp11/issues/455

dengert:
Yes looks #455 is the same problem.
With OpenSSL 3.0.5  it appears that `./ssl/engine` has been  added as `./ssl/engine-3` and libp11 will install the `libpkcs11.so` as a symlink to `pkcs11.so` 

So can both of you try modifying the instructions found in [README.md](https://github.com/OpenSC/libp11#using-the-engine-from-the-command-line) to use the engine-3 directory rather then the engine directory.
i.e. should have `dynamic_path = /usr/lib/ssl/engines-3/libpkcs11.so`


[...]

PR #457, opened 2022-07-29
Deffer initializing crypto routines in PKCS11 engine until needed

dengert:
Fixes:#456

bind_helper in eng_font.c is split into bind_helper and bind_helper2
The calls to ENGINE_set_RSA, ENGINE_set_EC, ENGINE_set_ECDH and
ENGINE_set_pkey_meths are moved to bind_helper2.

bind_helper2 is called from load_pubkey and load_privkey.

This in effect gets around the problem OpenSSL 3.0.x has when
it loads the pkcs11 engine from openssl.cnf, and then tries to use it
as a default provider even when no engine was specified on
the command line.

 On branch deffer_init_crypto
 Changes to be committed:
	modified:   eng_front.c

This is a stop gap fix, as engines are deprecated and OpenSSL is 
talking about their own PKCS11 provider. 

The following tests were run using OpenSSL-1.1.1q and OpenSSL-3.0.5:
Tested with commands that do not need the engine:
	 ./openssl ecparam -name brainpoolP256r1 -genkey -out t1.key
	 ./openssl req -subj /CN=ectest -new -x509 -key t1.key -out t1.crt

Tested with commands that use the PKCS11 engine:
         ./openssl req -subj /CN=ectest -new -x509 -key "slot_0-id_01" -engine pkcs11 -keyform engine -out /tmp/cert.pem
Two NIST PIV demo cards were uses. One with RSA and one with a EC prime256v1 key.

### DIFF
--- a/src/eng_front.c
+++ b/src/eng_front.c
@@ -82,7 +82,7 @@ static const ENGINE_CMD_DEFN engine_cmd_defns[] = {
 	{0, NULL, NULL, 0}
 };
 
-static int bind_helper2(ENGINE *e);
+static int bind_helper_methods(ENGINE *e);
 
 static ENGINE_CTX *get_ctx(ENGINE *engine)
 {
@@ -176,7 +176,7 @@ static EVP_PKEY *load_pubkey(ENGINE *engine, const char *s_key_id,
 	ctx = get_ctx(engine);
 	if (!ctx)
 		return 0;
-	bind_helper2(engine);
+	bind_helper_methods(engine);
 	return ctx_load_pubkey(ctx, s_key_id, ui_method, callback_data);
 }
 
@@ -189,7 +189,7 @@ static EVP_PKEY *load_privkey(ENGINE *engine, const char *s_key_id,
 	ctx = get_ctx(engine);
 	if (!ctx)
 		return 0;
-	bind_helper2(engine);
+	bind_helper_methods(engine);
 	pkey = ctx_load_privkey(ctx, s_key_id, ui_method, callback_data);
 #ifdef EVP_F_EVP_PKEY_SET1_ENGINE
 	/* EVP_PKEY_set1_engine() is required for OpenSSL 1.1.x,
@@ -223,7 +223,6 @@ static int bind_helper(ENGINE *e)
 			!ENGINE_set_ctrl_function(e, engine_ctrl) ||
 			!ENGINE_set_cmd_defns(e, engine_cmd_defns) ||
 			!ENGINE_set_name(e, PKCS11_ENGINE_NAME) ||
-
 			!ENGINE_set_load_pubkey_function(e, load_pubkey) ||
 			!ENGINE_set_load_privkey_function(e, load_privkey)) {
 		return 0;
@@ -239,7 +238,7 @@ static int bind_helper(ENGINE *e)
  * only add engine routines after a call to load keys
  */
 
-static int bind_helper2(ENGINE *e)
+static int bind_helper_methods(ENGINE *e)
 {
 	if (
 #ifndef OPENSSL_NO_RSA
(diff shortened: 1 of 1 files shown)


src/eng_front.c
```diff
@@ -219,6 +223,25 @@ static int bind_helper(ENGINE *e)
...
+	}
+}
+
+/*
+ * With OpenSSL 3.x, engines might be used because defined in openssl.cnf
+ * which will cause problems
+ * only add engine routines after a call to load keys
+ */
+
+static int bind_helper2(ENGINE *e)
```
Jakuje, line 242: technically, the code looks ok, but I am not a fan of `name2` function names as they do not give any hing what is the function about. Can we invent some better name, for example `bind_helper_methods` or something like that?

src/eng_front.c
```diff
@@ -219,6 +223,25 @@ static int bind_helper(ENGINE *e)
 			!ENGINE_set_ctrl_function(e, engine_ctrl) ||
 			!ENGINE_set_cmd_defns(e, engine_cmd_defns) ||
 			!ENGINE_set_name(e, PKCS11_ENGINE_NAME) ||
+
```
Jakuje, line 226: bogus empty line?

src/eng_front.c
dengert: OK.

src/eng_front.c
dengert: OK

mtrojnar:
It would be great to have a test for the pkcs11 engine defined in openssl.cnf...

mtrojnar:
This PR breaks our tests, so I reverted it in https://github.com/OpenSC/libp11/commit/e89aa1474ebe1cc6a425b4381ea15779b7343927.
Remember to always execute `make check` before submitting a commit.

dengert:
Do you have the output of what failed? 
Is `make check` run when a PR is submitted or committed? 

When running `make check` one problem I see is the all the `tests/pkcs11*.softhsm` scripts assume the  `openssl`  program is in the path and so it picks up the system version. 

For testing I build multiple versions of OpenSSL and install into `/opt/ossl-*/` and then build multiple versions of OpenSC and libp11 and install them in these same /opt/ossl-*/`  So even with build of 3.0.5, the openssl version used was  OpenSSL 1.1.1f  31 Mar 2020 on the Ubuntu 20.04 test system. 
I do have `/usr/lib/x86_64-linux-gnu/softhsm/libsofthsm2.so` but have not used it in a few years. 

The best I can tell right now, is a call to `bind_helper_methods` may need to be added to `ctx_ctrl_load_cert` and maybe other places.

mtrojnar:
@dengert Are you saying that `make check` works for you and have problems reproducing the errors? The two tests I listed in the revert commit consistently fail on all the platforms I tried (Arch, Ubuntu, macOS).

There is a possibility that our tests are defective. If you have a way to improve them then don't hesitate to do it.

Good news is that I just added executing tests with GitHub Actions today, so you will have your future PRs automatically tested.

dengert:
No it does not work for me, and the two routines that fail are the same you see:
 - pkcs11-uri-without-token.softhsm
 - search-all-matching-tokens.softhsm
 
In both they use `#!/bin/sh` which failed with a `if [[ ` statement to test OSTYPE. I have gotten further with changing to `#!/bin/bash`.

I have also set my PATH  for each version so correct openssl is found.
(Will look at at better way to do this.)

Both 3.0.5 and 1.1.1q are failing.  Will look at using gdb next.

dengert:
See #460 which passes tests when I run it. 

> Good news is that I just added executing tests with GitHub Actions today, so you will have your future PRs automatically tested.

But the GitHub action fails, but does not list the `test-suite.log` which would say where it was failing.

mtrojnar:
Good point.  I fixed the CI: https://github.com/OpenSC/libp11/commit/e2501dae948f45bbadddf39040142371486177bb

PR #460, opened 2022-08-06
Deffer init crypto2

dengert:
This is a replacement for `deffer_init_crypto` which was reverted. An additional commit was added to fix the use of $OSTYPE in `pkcs11-uri-without-token.softhsm` and `pkcs11-uri-without-token.softhsm` that were failing on Ubuntu where `/bin/sh -> /bin/dash`.   `$OSTYPE` in a shell variable and different shells, even on the same system, may or may not be define it. i.e. in`/bin/dash` or it may be different string.  If it is not defined the line in the scripts: `if [[ "${OSTYPE}" == "darwin"* ]]; then` may fail.  

A `case ... esac` was used instead which should work on any shell on any system even if not defined. 

This PR need to be tested on `MacOS`

The code was tested on Ubuntu 20.04 with OpenSSL-1.1.1q  where OpenSSL and libp11 were built and installed into /opt/ossl-1.1.1q.  
It was also tested on Ubuntu-22.04 with OpenSSL-3.0.2 which where installed by Ubuntu into /usr.  

The test scripts  where written to expect the openssl command to be in the system defined location. When the tests were run on Ubuntu 20.04  the PATH was exported so ` /opt/ossl-1.1.1q/bin/openssl` would be used.

### DIFF
--- a/src/eng_front.c
+++ b/src/eng_front.c
@@ -82,7 +82,7 @@ static const ENGINE_CMD_DEFN engine_cmd_defns[] = {
 	{0, NULL, NULL, 0}
 };
 
-static int bind_helper2(ENGINE *e);
+static int bind_helper_methods(ENGINE *e);
 
 static ENGINE_CTX *get_ctx(ENGINE *engine)
 {
@@ -176,7 +176,7 @@ static EVP_PKEY *load_pubkey(ENGINE *engine, const char *s_key_id,
 	ctx = get_ctx(engine);
 	if (!ctx)
 		return 0;
-	bind_helper2(engine);
+	bind_helper_methods(engine);
 	return ctx_load_pubkey(ctx, s_key_id, ui_method, callback_data);
 }
 
@@ -189,7 +189,7 @@ static EVP_PKEY *load_privkey(ENGINE *engine, const char *s_key_id,
 	ctx = get_ctx(engine);
 	if (!ctx)
 		return 0;
-	bind_helper2(engine);
+	bind_helper_methods(engine);
 	pkey = ctx_load_privkey(ctx, s_key_id, ui_method, callback_data);
 #ifdef EVP_F_EVP_PKEY_SET1_ENGINE
 	/* EVP_PKEY_set1_engine() is required for OpenSSL 1.1.x,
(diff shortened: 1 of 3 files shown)


mtrojnar:
Merge the master branch to see additional CI output details (added in https://github.com/OpenSC/libp11/commit/e2501dae948f45bbadddf39040142371486177bb).

mtrojnar:
I forked your repository and it seems to fix the tests.  I'll try merging it.

mtrojnar:
I hope I fixed it.  It was a real minefield: a change fixing three testing platforms was breaking a fourth platform.

mtrojnar:
Please take a look: https://github.com/OpenSC/libp11/commit/feb22a666ca361adb6f454bcb541281f8e9615f8

dengert:
Good. I was trying to submit a defferinit_crypto3 which was basically the same thing.  but with  `pkcs11-uri-without-tolen.softhsm`
```
# These URIs don't contain the token specification
-PRIVATE_KEY="pkcs11:object=server-key;type=private;pin-value=1234"
-PUBLIC_KEY="pkcs11:object=server-key;type=public;pin-value=1234"
+# Without id, tests fails as there are two slots,one good one uninitilized
+PRIVATE_KEY="pkcs11:id=%01%02%03%04;object=server-key;type=private;pin-value=1234"
+PUBLIC_KEY="pkcs11:id=%01%02%03%04;object=server-key;type=public;pin-value=1234"
```

gdb was showing a failed login, with a token with two slots, one good and one uninitialized. I suspect that is why the test was failing as it was trying to login to the uninitialized slot.

dengert:
https://github.com/OpenSC/libp11/commit/feb22a666ca361adb6f454bcb541281f8e9615f8 loos good.

mtrojnar:
Can you add a test with a pkcs11 engine defined in openssl.cnf? OpenSSL 3.0 seems to load our engine quite early, before its own infrastructure is fully initialized.  It would be good to have a test to detect any regressions.

dengert:
I am not very good at creating tests. But it looks like all the tests use the systems version of OpenSSL. Ubuntu-22.04 installs OpenSSL-3.0.2. 
So one way to do the test is to add an Ubuntu-22.04 to the build list of systems.

mtrojnar:
> But it looks like all the tests use the systems version of OpenSSL.

Only Linux tests use the OpenSSL provided by the OS.  Anyway, testing different versions of Ubuntu is a good idea: https://github.com/OpenSC/libp11/commit/a7b32b8b98824cba2154d387850dc14304a2b66b

PR #465, opened 2022-08-10
Add 'ec-ecparam.softhsm' test

ulrichb:
This adds an additional test for #456.

It runs `openssl ecparam -genkey ...` with an additional `-engine pkcs11`.

The test passes after @dengert's fixes for #456 (commit 5ab3c0debf2955c4cb942bf00f941c18e532b884) ... but not on the current master.

The commit which breaks the test is feb22a666ca361adb6f454bcb541281f8e9615f8. When reverting this commit on the current master the test becomes green (and all other are green too).


mtrojnar:
It needs `chmod a+x tests/ec-ecparam.softhsm`.

ulrichb:
Ooops, committed it under Windows. Fixed.

ulrichb:
... now they run and - yipee - the `ec-ecparam.softhsm` test fails. And as expected only in the OpenSSL 3 builds.

mtrojnar:
Thank you.  This test will be very useful for fixing it and and for finding any future regressions.

PR #474, opened 2022-09-08
Introduce keypair generation as engine ctrl command

istepic:
As discussed in https://github.com/OpenSC/libp11/pull/379 and
https://github.com/OpenSC/libp11/pull/378 we need a generic interface
that supports multiple algorithms for key generation. Attempt was made
to create a new keygen method and register it in PKCS11_pkey_meths() in
p11_pkey.c (so that it's possible to generate keys using OpenSSL's
EVP_PKEY_* API) but multiple design issues appeared. How and where do you
pass the key ID, token label and alike was the first question. As
suggested by the maintainer here:
https://github.com/OpenSC/libp11/pull/379#issuecomment-820588833,
app_data from EVP_PKEY_CTX was (mis)used and that worked well. The
reason why this approach was abandoned is because a good (or bad) way
to get a handle of the PKCS11_CTX_private, that is necessary for the
Cryptoki call, was not found.
The way other operations work is that they rely on the key being
loaded *_first_* through ENGINE_load_public(private)_key because this
is when the PKCS11_CTX gets initialized and a handle to
PKCS11_OBJECT_private gets set to the ex_data of the underlying key.
Key generation obviously cannot rely on that mechanism since key
doesn't yet exist.

Instead, a generic PKCS11_generate_key interface was made that
takes a structure describing the key generation algorithm. For now
it only contains simple options like curve name for ECC or number
of bits for RSA key generation. This interface can then be used
as any other PKCS11 wrapper interface or using the ENGINE control
commands. Using it with ENGINE control commands is demonstrated in
the new tests/keygen.c file.

Code for ECC keygen was taken from:
https://github.com/OpenSC/libp11/pull/379 and reworked to compile and
work with some new additions to libp11 i.e. templates.

### DIFF
--- a/src/eng_back.c
+++ b/src/eng_back.c
@@ -257,7 +257,43 @@ static int ENGINE_CTX_ctrl_set_vlog(ENGINE_CTX *ctx, void *cb)
 	return 1;
 }
 
-int ENGINE_CTX_ctrl(ENGINE_CTX *ctx, int cmd, long i, void *p, void (*f)(void))
+static int ENGINE_CTX_keygen(ENGINE_CTX *ctx, void *p)
+{
+	int rv;
+	PKCS11_KGEN_ATTRS *kg_attrs = p;
+	PKCS11_SLOT* found_slot = NULL;
+
+	if (kg_attrs == NULL)
+		return 0;
+
+	/* Delayed libp11 initialization */
+	if (UTIL_CTX_init_libp11(ctx->util_ctx)) {
+		ENGerr(ENG_F_CTX_LOAD_OBJECT, ENG_R_INVALID_PARAMETER);
+		return 0;
+	}
+
+	found_slot = UTIL_CTX_find_token(ctx->util_ctx, kg_attrs->token_label);
+	if (!found_slot)
+		return 0;
+
+	/* Try logging in */
+	ERR_clear_error();
+	if (!(found_slot->token->loginRequired && UTIL_CTX_login(ctx->util_ctx,
+	    found_slot, ctx->ui_method, ctx->ui_data)))
+		return 0;
+
+	rv = PKCS11_keygen(found_slot->token, kg_attrs);
+	if (rv < 0) {
+		ENGINE_CTX_log(ctx, LOG_ERR,
+			"Failed to generate a key pair on the token. Error code: %d\n",
+			rv);
+		return 0;
+	}
+
+	return 1;
+}
+
+int ENGINE_CTX_ctrl(ENGINE_CTX *ctx, int cmd, long i, void *p, void (*f)())
 {
 	(void)i; /* We don't currently take integer parameters */
 	(void)f; /* We don't currently take callback parameters */
@@ -293,6 +329,8 @@ int ENGINE_CTX_ctrl(ENGINE_CTX *ctx, int cmd, long i, void *p, void (*f)(void))
 		return ENGINE_CTX_ctrl_set_vlog(ctx, p);
 	case CMD_DEBUG_LEVEL:
 		return ENGINE_CTX_ctrl_set_debug_level(ctx, (int)i);
+	case CMD_KEYGEN:
+		return ENGINE_CTX_keygen(ctx, p);
 	default:
 		ENGerr(ENG_F_CTX_ENGINE_CTRL, ENG_R_UNKNOWN_COMMAND);
 		break;
--- a/src/libp11-int.h
+++ b/src/libp11-int.h
@@ -332,9 +332,13 @@ extern int pkcs11_generate_random(PKCS11_SLOT_private *, unsigned char *r, unsig
 /* Internal implementation of deprecated features */
 
 /* Generate and store a private key on the token */
-extern int pkcs11_generate_key(PKCS11_SLOT_private *tpriv,
-	int algorithm, unsigned int bits,
-	char *label, unsigned char *id, size_t id_len);
+extern int pkcs11_rsa_keygen(PKCS11_SLOT_private *tpriv,
+	unsigned int bits, const char *label, const unsigned char *id,
+	size_t id_len, const PKCS11_params* params);
+
+extern int pkcs11_ec_keygen(PKCS11_SLOT_private *tpriv,
+	const char *curve , const char *label, const unsigned char *id,
+	size_t id_len, const PKCS11_params* params);
 
 /* Get the RSA key modulus size (in bytes) */
 extern int pkcs11_get_key_size(PKCS11_OBJECT_private *);
--- a/src/p11_key.c
+++ b/src/p11_key.c
@@ -29,13 +29,19 @@
 /* The maximum length of PIN */
 #define MAX_PIN_LENGTH   256
 
-static int pkcs11_find_keys(PKCS11_SLOT_private *, CK_SESSION_HANDLE, unsigned int, PKCS11_TEMPLATE *);
-static int pkcs11_next_key(PKCS11_CTX_private *ctx, PKCS11_SLOT_private *,
-	CK_SESSION_HANDLE session, CK_OBJECT_CLASS type);
+static int pkcs11_find_keys(PKCS11_SLOT_private *, CK_SESSION_HANDLE, unsigned int,
+	PKCS11_TEMPLATE *);
 static int pkcs11_init_key(PKCS11_SLOT_private *, CK_SESSION_HANDLE session,
 	CK_OBJECT_HANDLE o, CK_OBJECT_CLASS type, PKCS11_KEY **);
+static int pkcs11_init_keygen(PKCS11_SLOT_private *, CK_SESSION_HANDLE *);
+static int pkcs11_next_key(PKCS11_CTX_private *ctx, PKCS11_SLOT_private *,
+	CK_SESSION_HANDLE session, CK_OBJECT_CLASS type);
 static int pkcs11_store_key(PKCS11_SLOT_private *, EVP_PKEY *, CK_OBJECT_CLASS,
 	char *, unsigned char *, size_t, PKCS11_KEY **);
+static void pkcs11_common_pubkey_attr(PKCS11_TEMPLATE *, const char *,
+	const unsigned char *, size_t);
+static void pkcs11_common_privkey_attr(PKCS11_TEMPLATE *, const char *,
+	const unsigned char *, size_t, const PKCS11_params *);
 
 /* Helper to acquire object handle from given template */
 static CK_OBJECT_HANDLE pkcs11_handle_from_template(PKCS11_SLOT_private *slot,
@@ -272,11 +278,12 @@ int pkcs11_reload_object(PKCS11_OBJECT_private *obj)
 }
 
 /**
- * Generate a key pair directly on token
+ * Generate RSA key pair directly on token
  */
-int pkcs11_generate_key(PKCS11_SLOT_private *slot, int algorithm, unsigned int bits,
-		char *label, unsigned char *id, size_t id_len) {
-
+int pkcs11_rsa_keygen(PKCS11_SLOT_private *slot, unsigned int bits,
+		const char *label, const unsigned char *id, size_t id_len,
+		const PKCS11_params *params)
+{
 	PKCS11_CTX_private *ctx = slot->ctx;
 	CK_SESSION_HANDLE session;
 	PKCS11_TEMPLATE pubtmpl = {0}, privtmpl = {0};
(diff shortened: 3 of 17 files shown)


istepic:
Hello @mtrojnar, does this approach to key generation seem plausible _at all_ to you?

ldts:
what is the status of this PR? has it been abandoned as well? It seems there was a previous attempt at implementing EC keygen which didn't prosper either (https://github.com/OpenSC/libp11/pull/379).

is there are reason not to pursue either of them?

dengert:
> what is the status of this PR? has it been abandoned as well? It seems there was a previous attempt at implementing EC keygen which didn't prosper either (https://github.com/OpenSC/libp11/pull/379).

is there are reason not to pursue either of them?

IMHO biggest issue with both of these is when will OpenSSL drops engine support and what happens to libp11.

mwasilew:
Is there an alternative way of generating EC key pairs on the token? It is currently possible to generate EC keys with pkcs11-tool, but not libp11 API.

istepic:
> Is there an alternative way of generating EC key pairs on the token? It is currently possible to generate EC keys with pkcs11-tool, but not libp11 API.

No alternative way for EC keys.

mtrojnar:
The engine interface looks good to me.

On the other hand, changing the definition of `PKCS11_generate_key()` is unacceptable, because it breaks each and every existing function that uses that API function.  You need to define a new API function instead (e.g., `PKCS11_keygen()`) and reimplement the existing `PKCS11_generate_key()` API function as a wrapper for the new `PKCS11_keygen()`.

rafajunio:
Hello @mtrojnar , can you please check if the PR is good for you now? 

I requested Ivan access for change the patch.

I do not known what to do with the hex_to_bin in the p11_front.c, but I do not want to introduce even more changes to the PR.

mtrojnar:
@rafajunio

The API looks better now.  I think it should be possible to merge this feature.

@olszomal and myself are in the middle of implementing PKCS#11 provider. More internal changes in PIN UI will be implemented soon.  Please wait a few weeks for the code changes to stabilize, and then I'll ask you to rebase your PR on top of the new code.  What do you think?

rafajunio:
@mtrojnar 
I tried to refactor as much as possible, I also tried to create the provider test already but I could not. Either way I upload the code for you start the review, if you have some hint please let me known

tests/keygen.c
```diff
@@ -0,0 +1,244 @@
...
+		printf("RSA Sign-verify failed with err code: %d\n", ret);
+		goto cleanup;
+	}
+	printf("RSA Sign-verify success\n");
+
+	ret = 0;
+cleanup:
+	if (!engine)
+		ENGINE_finish(engine);
+
```
olszomal, line 242: Remember to call EVP_PKEY_free()

tests/keygen.c
```diff
@@ -0,0 +1,244 @@
...
+	if (!ENGINE_init(engine)) {
+		printf("Could not initialize engine\n");
+		display_openssl_errors(__LINE__);
+		goto cleanup;
+	}
+
+	/*
+	 * EC key generation test
+	 */
+	PKCS11_EC_KGEN ec = {
```
olszomal, line 177: Declare structures and variables at the beginning of the code block, following the C89/C90 standard.

tests/keygen.c
```diff
@@ -0,0 +1,244 @@
...
+	EVP_PKEY *ecpb = ENGINE_load_public_key(engine, "1234", NULL, NULL);
+	EVP_PKEY *ecpr = ENGINE_load_private_key(engine, "1234", NULL, NULL);
+	if ((ret = sign_verify_test(ecpr, ecpb)) < 0) {
+		printf("ECC Sign-verify failed with err code: %d\n", ret);
+		goto cleanup;
+	}
+	printf("ECC Sign-verify success\n");
+
+	/*
+	 * RSA key generation test
```
olszomal, line 209: Split the test into two separate cases for RSA and EC key generation to maintain consistency with existing tests.

src/eng_back.c
```diff
@@ -257,7 +257,54 @@ static int ENGINE_CTX_ctrl_set_vlog(ENGINE_CTX *ctx, void *cb)
...
+	 * continue and try to login first
+	 */
+	if (!UTIL_CTX_get_force_login(ctx->util_ctx)) {
+		ERR_clear_error();
+		if (!PKCS11_keygen(found_slot->token, kg_attrs)) {
+			return 1;
+		}
+	}
+
+	// Try with logging in
```
olszomal, line 290: This project uses C-style comments instead of C++.

tests/keygen-prov.c
```diff
@@ -0,0 +1,299 @@
...
+	if (argc < 4) {
+		printf("Too few arguments\n");
+		printf("%s [TOKEN1] [KEY-LABEL] [PIN] [module]\n", argv[0]);
+		exit(ret);
+	}
+	const char *token_label = argv[1], *key_label = argv[2], *module = argv[4];
+	char *key_pass = argv[3];
+
+	/* Load pkcs11prov and default providers */
+	if (!providers_load()) {
```
olszomal, line 150: Unfortunately, the OpenSSL provider currently only supports retrieving objects from the token, meaning it cannot be used for key generation.
Please remove the key generation test for the provider.

src/util.h
```diff
@@ -61,9 +61,13 @@ void UTIL_CTX_free_libp11(UTIL_CTX *ctx);
 void UTIL_CTX_set_vlog_a(UTIL_CTX *ctx, PKCS11_VLOG_A_CB vlog);
 void UTIL_CTX_set_debug_level(UTIL_CTX *ctx, int debug_level);
 void UTIL_CTX_log(UTIL_CTX *ctx, int level, const char *format, ...);
+int UTIL_CTX_hex_to_bin(UTIL_CTX *ctx, const char *in, char *out, size_t *outlen);
 
 int UTIL_CTX_set_pin(UTIL_CTX *ctx, const char *pin);
 void UTIL_CTX_set_force_login(UTIL_CTX *ctx, int force_login);
+int UTIL_CTX_get_force_login(UTIL_CTX *ctx);
```
olszomal, line 68: Remove the unused function UTIL_CTX_get_force_login().

olszomal (changes requested):
Excellent work – the code is clean and well-structured, making it easy to review.

I confirm that key pair generation works correctly for both the libp11 library and the engine.

Unfortunately, the OpenSSL provider currently only supports retrieving objects from the token, meaning it cannot be used for key generation. Please remove the key generation test for the provider.

src/eng_back.c
```diff
@@ -257,7 +257,43 @@ static int ENGINE_CTX_ctrl_set_vlog(ENGINE_CTX *ctx, void *cb)
...
+		return 0;
+	}
+
+	found_slot = UTIL_CTX_find_token(ctx->util_ctx, kg_attrs->token_label);
+	if (!found_slot)
+		return 0;
+
+	/* Try with logging in */
+	ERR_clear_error();
+	if (found_slot->token->loginRequired && UTIL_CTX_login(ctx->util_ctx,
```
olszomal, line 281: If token login is required, UTIL_CTX_login() must be called. However, try to generate a key pair.

src/p11_key.c
```diff
@@ -332,6 +326,82 @@ int pkcs11_generate_key(PKCS11_SLOT_private *slot, int algorithm, unsigned int b
...
+	if (curve_nid == NID_undef)
+		curve_nid = OBJ_sn2nid(curve);
+	if (curve_nid == NID_undef)
+		curve_nid = OBJ_ln2nid(curve);
+	if (curve_nid == NID_undef)
+		return -1;
+	curve_obj = OBJ_nid2obj(curve_nid);
+	if (!curve_obj)
+		return -1;
+	// convert to DER format and take just the length
```
olszomal, line 363: Use C-style comments instead of C++. This note applies to all comments; sorry for the confusion.

tests/Makefile.am
```diff
@@ -72,6 +76,7 @@ rsa_pss_sign_prov_SOURCES = rsa-pss-sign-prov.c helpers_prov.c
 rsa_oaep_prov_SOURCES = rsa-oaep-prov.c helpers_prov.c
 store_cert_prov_SOURCES = store-cert-prov.c helpers_prov.c
 check_all_prov_SOURCES = check-all-prov.c helpers_prov.c
+keygen_prov_SOURCES = keygen-prov.c helpers_prov.c
```
olszomal, line 79: Unused

tests/ec-keygen.c
```diff
@@ -0,0 +1,229 @@
...
+	}
+	printf("ECC keys generated\n");
+
+	ecpb = ENGINE_load_public_key(engine, "2233", NULL, NULL);
+	ecpr = ENGINE_load_private_key(engine, "2233", NULL, NULL);
+	if ((ret = sign_verify_test(ecpr, ecpb)) < 0) {
+		printf("ECC Sign-verify failed with err code: %d\n", ret);
+		goto cleanup;
+	}
+	printf("ECC Sign-verify success\n");
```
olszomal, line 218: Formally, it should be EC keys instead of ECC keys, as EC refers specifically to Elliptic Curve keys, while ECC refers to the broader field of Elliptic Curve Cryptography.

src/libp11-int.h
```diff
@@ -332,9 +332,13 @@ extern int pkcs11_generate_random(PKCS11_SLOT_private *, unsigned char *r, unsig
 /* Internal implementation of deprecated features */
 
 /* Generate and store a private key on the token */
-extern int pkcs11_generate_key(PKCS11_SLOT_private *tpriv,
-	int algorithm, unsigned int bits,
-	char *label, unsigned char *id, size_t id_len);
+extern int pkcs11_rsa_keygen(PKCS11_SLOT_private *tpriv,
+	unsigned int bits, const char *label, const unsigned char* id,
+	size_t id_len, const PKCS11_params* params);
```
olszomal, line 337: Please use `TYPE *ptr` instead of `TYPE* ptr` to follow the project's convention.

olszomal (changes requested):
Generally well done! Some minor style tweaks, but also a significant issue that needs attention.

tests/ec-keygen.c
```diff
@@ -0,0 +1,229 @@
...
+	ecpr = ENGINE_load_private_key(engine, "2233", NULL, NULL);
+	if ((ret = sign_verify_test(ecpr, ecpb)) < 0) {
+		printf("EC Sign-verify failed with err code: %d\n", ret);
+		goto cleanup;
+	}
+	printf("EC Sign-verify success\n");
+
+	ret = 0;
+cleanup:
+	if (!engine)
```
olszomal, line 222: This condition is incorrect.
ENGINE_finish(engine) should be called unconditionally to release the functional reference obtained via ENGINE_init(). Please remove the if statement and call ENGINE_finish(engine); directly. Skipping it when engine == NULL is safe, as the function handles that internally.

tests/rsa-keygen.c
```diff
@@ -0,0 +1,229 @@
...
+	rsapr = ENGINE_load_private_key(engine, "4321", NULL, NULL);
+	if ((ret = sign_verify_test(rsapr, rsapb)) < 0) {
+		printf("RSA Sign-verify failed with err code: %d\n", ret);
+		goto cleanup;
+	}
+	printf("RSA Sign-verify success\n");
+
+	ret = 0;
+cleanup:
+	if (!engine)
```
olszomal, line 222: The condition is incorrect.

olszomal (changes requested):

olszomal (approved):
Looks good — thanks for the changes!
I’ll take care of the small style issue separately to keep things moving.

src/p11_key.c
```diff
@@ -272,48 +278,36 @@ int pkcs11_reload_object(PKCS11_OBJECT_private *obj)
...
+{
 	PKCS11_CTX_private *ctx = slot->ctx;
 	CK_SESSION_HANDLE session;
 	PKCS11_TEMPLATE pubtmpl = {0}, privtmpl = {0};
 	CK_MECHANISM mechanism = {
 		CKM_RSA_PKCS_KEY_PAIR_GEN, NULL_PTR, 0
 	};
 	CK_ULONG num_bits = bits;
-	CK_BYTE public_exponent[] = { 1, 0, 1 };
+	CK_BYTE public_exponent[] = { 1, 0, 0, 0, 1 };
```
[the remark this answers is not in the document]
proxyrer: Interesting! So the value is indeed a valid exponent? Sorry, I'm not too well-versed in cryptography. It seems then that the real limitation that got me here is that the software I use that signs the CSR does not support large public key exponents. Is there a way to specify desired public key exponent in libp11 during key generation? Also let me know if you'd rather I open an issue to move this discussion elsewhere.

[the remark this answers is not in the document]
mtrojnar: @istepic Would you kindly share your rationale for changing `{ 1, 0, 1 }` to `{ 1, 0, 0, 0, 1 }`?

PR #495, opened 2023-04-16
slot: fix token initialization

ldts:
The current interface accepts a string as the token label.

The specification however specifies that the label must point to a 32-byte memory location, which MUST be padded with blank characters and which MUST not be null-terminated.

This fix allows using libp11 with libtpm2_pkcs11 since the TPM pkcs#11 implementation does enforce the requirement (libsofthsm does not).


ldts:
these CI errors seem unrelated to the PR

ldts:
Could you pick this fix for your next release please? We'd rather not carry the patch.

mtrojnar:
I'm afraid `{' '}` doesn't do what you think it does:
```
$ cat initializer.c 
#include <stdio.h>

int main() {
    unsigned char ck_label[32] = {' '};
    int i;
    printf("ck_label:");
    for(i=0; i<sizeof ck_label; ++i)
        printf(" %d", ck_label[i]);
    printf("\n");
    return 0;
}
$ gcc -o initializer -Wall initializer.c 
$ ./initializer 
ck_label: 32 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
```
Please use memset(3) instead.

mtrojnar:
Argh.  You didn't even check whether your commit can be built...

ldts:
> Argh. You didn't even check whether your commit can be built...

dude, what are you talking about

ldts:
> Argh. You didn't even check whether your commit can be built...

right, apologies. I expected the CI to do this on my behalf.

PR #503, opened 2023-06-06
Detect openSSL 3.1; compatible to openSSL 3.0

DimStar77:


floppym:
So what happens when OpenSSL 3.2 is released?

Issue #518, opened 2023-09-25
PKCS11_store_certificate generates CKR_ATTRIBUTE_TYPE_INVALID error


mihaicristiantanase:
I'm using the latest version of the library (master: 6c96847f1f52a5ccc76e8f8d14820cc4d6af1ecb) to write a certificate on a cryptographic device.
The call to **PKCS11_store_certificate** function with an X509 object previously read from a PEM file (via **PEM_read_X509** function) always returns -1 and the certificate is not written.

I have investigated the problem further and found out that this line (in function pkcs11_store_certificate of p11_cert.c):

	rv = CRYPTOKI_call(ctx, C_CreateObject(session, tmpl.attrs, tmpl.nattr, &object));

returns a value of 18 (0x12) which, in the PKCS#11 standard is translated to **CKR_ATTRIBUTE_TYPE_INVALID**.

The process of finding the culprit was tiring but I've managed to reduce the problem to those lines:

	pkcs11_addattr_var(&tmpl, CKA_NAME_HASH_ALGORITHM, ckm_md);
	if (X509_pubkey_digest(x509,evp_md,md,&md_len))
		pkcs11_addattr(&tmpl, CKA_HASH_OF_SUBJECT_PUBLIC_KEY, md, md_len);

If those two attributes (**CKA_NAME_HASH_ALGORITHM** and **CKA_HASH_OF_SUBJECT_PUBLIC_KEY**) are not set, then the certificate is successfully written.
If I keep the "**CKA_NAME_HASH_ALGORITHM**" attribute, then **CRYPTOKI_call** returns the same error (18).
On the other hand, if I keep the "**CKA_HASH_OF_SUBJECT_PUBLIC_KEY**" attribute, then **CRYPTOKI_call** returns error 5 (**CKR_GENERAL_ERROR**).

This behaviors is identical on two different devices:
1. Manufacturer "Athena Smartcard Solutions", model "IDProtect";
2. Manufacturer "Gemalto", model "ID Prime MD";

Consulting the standard (http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.html#_Toc416959712), it seems that this function sets up a wrong combination of attributes, that is, when **CKA_VALUE** is non-empty, **CKA_URL** must be empty, which permits ("can") **CKA_HASH_OF_SUBJECT_PUBLIC_KEY** to be empty.

My questions are:
[...]

PR #519, opened 2023-10-03
Fix for PKCS11_store_certificate generates CKR_ATTRIBUTE_TYPE_INVALID error

mihaicristiantanase:
This fixes the issue https://github.com/OpenSC/libp11/issues/518


dengert:
`C_GetInfo` is called when the module is loaded:

  https://github.com/OpenSC/libp11/blob/master/src/p11_load.c#L94-L107

The version number could be save once in `PKCS11_CTX`, rather then calling C_GetInfo each time the library version is needed.
Personally, I would save it as a `CK_VERSION library_version` structure.  Or could be saved as `major * 1000 + minor` as you are doing.  
There may be other places in libp11 that might need the version number too.

@mtrojnar  comments?

mtrojnar:
I fully support @dengert's suggestions.

Also, please do not add another source file for two trivial functions.  As @dengert pointed out, it's one PKCS11 API call after the module is loaded and initialized, and one field in the context structure, so no new functions seem to be needed.

mihaicristiantanase:
Thanks for the feedback.

First, saving the version at library load is a good idea. The only problem I see in reusing CK_VERSION is that I need to include "pkcs11.h" into "libp11.h", which is probably not a good idea. I'm thinking of creating a different type in "libp11.h" (ex: Version) that mirrors CK_VERSION. @dengert, what do you think?

Second, I don't see the issue with adding a new source file. Even if a source file contains trivial content, you don't really want to repeat it in all places it is used. Probably I'm missing something here...
Anyway, @mtrojnar, where do you suggest I should add version related logic?

mtrojnar:
Adding https://github.com/OpenSC/libp11/commit/9ac6919996beb76361c2d00faec1af7051c84c99 was easier than explaining.

Avoid converting them to an integer or adding an encapsulation layer.   An additional abstraction layer makes the code harder instead of simpler to understand.  Simply create and appropriate condition every time it's needed.

General advice: Use the [KISS principle](https://en.wikipedia.org/wiki/KISS_principle).

mihaicristiantanase:
https://github.com/OpenSC/libp11/commit/9ac6919996beb76361c2d00faec1af7051c84c99 adds a new type to **PKCS11_CTX**. Any idea how can this be accessed from a **PKCS11_SLOT_private** object?

mtrojnar:
Good point.  It shouldn't go into the public API anyway.  Apply https://github.com/OpenSC/libp11/commit/c6e69c944dcedf5a57d6d7a363f83c023ba4813e and refer to `slot->ctx->cryptoki_version.major` and `slot->ctx->cryptoki_version.minor`.

mihaicristiantanase:
I've used the simpler **CK_VERSION** instead of **struct _CK_VERSION**. @mtrojnar, is there a reason you used the latter in **PKCS11_SLOT_private**?
Also, I've added a version comparison function in p11_cert.c instead of using a new source file. Probably, this will be moved to a separate place when the comparison is needed elsewhere.

mihaicristiantanase:
No problem. Thank you for merging the changes!

Issue #527, opened 2024-03-13
Library cleanup issues with OpenSSL - libp11 - yubihsm_pkcs11.so chain


pugo:
We use libp11 in a chain like the following: 

`CST - OpenSSL - libpkcs11.so - yubihsm_pkcs11.so - YubiHSM 2`

Where CST is i.MX Code Signing Tool, which is used to sign images for secure boot on NXP SOC:s. 

When stress testing our signing I saw that the PKCS11 sessions are not correctly released, which after a short while under load causes errors due to lack of free sessions. With the `yubihsm-connector` the maximum number of consecutive sessions is 16 and the default timeout for inactive sessions is 30 seconds. When running more frequent than the timeouts can cope with signing operations fail. 

I have been investigating the not released sessions, since we have a setup where signing operations can come in bursts that could potentially be hurt by the described problem. This could cause random build failures.

The results from my investigations are some smaller issues in the CST tool, but also one or two in libp11. They are described below.

_I will make a PR for the first one._

### Reference counting in p11_key.c `pkcs11_object_free()`

In the following code in `p11_key.c` the decrease of `obj->refcnt` happens before the following handling
of the `obj->evp_key`. 

```
void pkcs11_object_free(PKCS11_OBJECT_private *obj)
{
        if(!obj)
                return;

        if (pkcs11_atomic_add(&obj->refcnt, -1, &obj->lock) != 0)
                return;
        if (obj->evp_key) {
                /* When the EVP object is reference count goes to zero, 
                 * it will call this function again. */
                EVP_PKEY *pkey = obj->evp_key;
                obj->evp_key = NULL;
                EVP_PKEY_free(pkey);
                return;
        }
        pkcs11_slot_unref(obj->slot);
        X509_free(obj->x509);
        OPENSSL_free(obj->label);
        pthread_mutex_destroy(&obj->lock);
        OPENSSL_free(obj);
}
```

In the `if (obj->evp_key)` section there is a comment that states that the `pkcs11_object_free()` will
get a new call as a result of the `EVP_PKEY_free(pkey)` call. At the end of that scope it returns.

When the second call comes, as stated in the comment, the `obj->refcnt` will be decreased once more.
This means that it will likely go negative and not `0` as tested in the if-clause. This means that it will return
and the five clean-up lines at the end of the function will never be run.

_I will make a PR for the above._

### Workaround in `eng_front.c` blocks cleanup

[...]

PR #528, opened 2024-03-13
Decrement ref counter only once on key object free

pugo:
This is a suggested solution to the issue I describe in issue #527. 

By moving the decrement of `obj->refcnt` to _after_ the if-section for `obj->evp_key` the counter will not be decremented twice if `obj->evp_key` is `true` and we get another call as result of the `EVP_PKEY_free()`.

Without this patch the decrement will be done twice and the comparison with `0` will fail since the value goes negative. This means that the cleanup code at the end is never run.

As far as I can see there shouldn't be scenarios where the current behavior is wanted, but I of course do not have the big picture as clear as you have.

### DIFF
--- a/src/p11_key.c
+++ b/src/p11_key.c
@@ -185,8 +185,6 @@ void pkcs11_object_free(PKCS11_OBJECT_private *obj)
 	if(!obj)
 		return;
 
-	if (pkcs11_atomic_add(&obj->refcnt, -1, &obj->lock) != 0)
-		return;
 	if (obj->evp_key) {
 		/* When the EVP object is reference count goes to zero,
 		 * it will call this function again. */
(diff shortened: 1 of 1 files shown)


mtrojnar:
This issue was fixed in cec62fb7588dd7053c50b09913421cbd3e521c35:

```C
--- a/src/p11_key.c
+++ b/src/p11_key.c
@@ -187,14 +187,6 @@ void pkcs11_object_free(PKCS11_OBJECT_private *obj)
 
        if (pkcs11_atomic_add(&obj->refcnt, -1, &obj->lock) != 0)
                return;
-       if (obj->evp_key) {
-               /* When the EVP object is reference count goes to zero,
-                * it will call this function again. */
-               EVP_PKEY *pkey = obj->evp_key;
-               obj->evp_key = NULL;
-               EVP_PKEY_free(pkey);
-               return;
-       }
        pkcs11_slot_unref(obj->slot);
        X509_free(obj->x509);
        OPENSSL_free(obj->label);
```

Issue #541, opened 2024-09-02
openssl pkcs11 engine: a newly created key cannot be used if one with the same url was used before


sbrandt-aixtrusion:
We have the following issue:

We use openssl 3.0.13 to OP-TEE (arm embedded) via the PKCS11 engine (libp11 0.4.12)
We can create keys via pkcs11, use them from openssl, everything is okay.
But we also need to create new keys, when we create a new certificate enrollment.

Now, it appears that the openssl pkey is "cached" inside the engine. If we create a new pkcs11 key, the public part of the pkey is still cached and thus, the old key, while the private part is passed through pkcs11 to the op-tee implementation and uses, of course, the new key.
This makes the following workflow impossible:
* start the program
* use the pkcs11 key as EVP_PKEY
* works without issue
* free all instances of that EVP_PKEY
* create a new pkcs11 key (same name/url/id) via pkcs11-interface
* get a reference to the new pkcs11 key as EVP_PKEY
That EVP_PKEY is no longer able to verify its own signatures, as the public part is still the old key, while the private part is the new key.
An application restart fixes that problem, but that is not a solution that can be implemented in our project.

Looking into the code of the pkcs11 engine, I found that the created EVP_PKEY gets an EVP_PKEY_up_ref, and is returned with a reference count of 2, not of 1. So, when the user program frees "all references" of that EVP_PKEY, it is not actually freed. This of course explains why the new key still has the old public information, instead of retrieving the new public information from the new pkcs11 key. This is quite obviously not just a bug (very old code), but "by design", though not visibly explained.

When calling another EVP_PKEY_free, the key reference is of course freed completely. But it is no longer possible to get an EVP_PKEY reference. This reason for this is unclear, I have not yet debugged that far.

So, there are two "broken" flows.

1. during the runtime of a program, a pkcs11 engine key cannot be re-created and then used successfully by openssl, due to EVP_PKEY reference counting being used for caching the openssl/pkcs11 key "bridge". An application restart is needed.
2. without that key bridge, retrieving the same key again is impossible.

When we set up the project, Providers for PKCS11 where not yet sufficiently mature to be used. Switching everything to that would only make sense if we can be sure that this problem is not present there.

The issue cannot be easily reproduced in minimal sample code, as it requires a step outside of the application to create the new key. 

The attached sample code should reproduce the problem, assuming you have a working pkcs11 setup.

Output is like:
```
$ PIN=0000 /usr/bin/pkcs11-problem
[...]

PR #552, opened 2024-10-23
p11_key: do not assume a key handle relate to a previously loaded key

etienne-lms:
Do not assume the PKCS#11 handle value stored in libp11 key cache still relates to the very same object previously loaded. Maybe the client application has modified the PKCS#11 token content using an interface that does not go through the OpenSSL pkcs11 engine. This change addresses the issue by reloading the object attributes from the token when the PKCS#11 handle value is present in the engine cache.

### DIFF
--- a/src/p11_key.c
+++ b/src/p11_key.c
@@ -642,17 +642,28 @@ static int pkcs11_init_key(PKCS11_SLOT_private *slot, CK_SESSION_HANDLE session,
 {
 	PKCS11_keys *keys = (type == CKO_PRIVATE_KEY) ? &slot->prv : &slot->pub;
 	PKCS11_OBJECT_private *kpriv;
-	PKCS11_KEY *key, *tmp;
+	PKCS11_KEY *key = NULL, *tmp;
 	int i;
 
-	/* Prevent re-adding existing PKCS#11 object handles */
+	/*
+	 * Prevent re-adding and re-using existing PKCS#11 object handles.
+	 * If handle is found in cache, remove it a re-create a new OpenSSL
+	 * key for that handle. This prevents assuming the handle value still
+	 * relates to the very same object while client may be modified the
+	 * pkcs11 token data not through this pkcs11 engine.
+	 */
 	/* TODO: Rewrite the O(n) algorithm as O(log n),
 	 * or it may be too slow with a large number of keys */
 	for (i = 0; i < keys->num; ++i) {
-		if (PRIVKEY(&keys->keys[i])->object == object) {
-			if (ret)
-				*ret = &keys->keys[i];
-			return 0;
+		if (PRIVKEY(&keys->keys[i]) &&
+		    PRIVKEY(&keys->keys[i])->object == object) {
+			key = &keys->keys[i];
+
+			if (key->_private) {
+				pkcs11_object_free(PRIVKEY(key));
+				key->_private = NULL;
+			}
+			break;
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


mtrojnar:
The purpose of caching is precisely to reuse stored values. If I understand this PR correctly, it effectively disables caching because keys may have been modified by other means. The correct solution for #541 appears to be reloading the engine after adding, removing, or modifying keys. Keep in mind that closing and reopening the engine does not invalidate any `EVP_PKEY` objects acquired from the previous instance, as they retain a reference to that earlier engine.

etienne-lms:
Thanks @mtrojnar for your feedback.
Indeed, after a sequence like `eng = ENGINE_by_id("pkcs11"); ENGINE_free(eng); ENGINE_remove(eng);` 
the engine loaded with `ENGINE_by_id("pkcs11")` no more reproduces the issue related to pkcs11 engine caching object handle.

I'll close this P-R.

sbrandt-aixtrusion:
When I call ENGINE_remove, I can't reload the engine with ENGINE_by_id.

When I call ENGINE_finish and ENGINE_free, and then ENGINE_by_id, I get the same public key, so, engine was not unloaded.

With finish, remove and free:
```
pkcs11-problem 
destroyPkcs11Object: trying to delete 1 objects of label testkey, class 2
destroyPkcs11Object: trying to delete 1 objects of label testkey, class 3
Private and public key 'testkey' have been deleted by name.
Private and public key 'testkey' have been generated.
Trying to check key validity by signing and verifying.
Public key:
-----BEGIN PUBLIC KEY-----
MHYwEAYHKoZIzj0CAQYFK4EEACIDYgAEC6M6psEuLIdpxSZ4uR63WPndeqbzwBFx
RnvsGGYrXRGGvVLftxdpz5Bl03BcTXJ1oOXhwxuTmuuMzsii+mbMp0mOAyLnsFR9
gcw4/smW9Sk7VKeiXwn6wtTnFBUx/Bgq
-----END PUBLIC KEY-----

Signature created
Signature verified
First PKCS11 key is valid, and has been freed.
Reloading engine.
calling ENGINE_finish(0xaaab1675c710)
calling ENGINE_remove(0xaaab1675c710)
calling ENGINE_free(0xaaab1675c710)
Info: freed 'pkcs11' engine at 0xaaab1675c710
Caught exception: EVP_ENGINE_by_id(pkcs11) failed: - SSL error:1300006D:engine routines::init failed
- SSL error:13000074:engine routines::no such engine
```

Just free and remove (order changed):
```
...
Reloading engine.
calling ENGINE_free(0xaaaaf37c7710)
calling ENGINE_remove(0xaaaaf37c7710)
Info: freed 'pkcs11' engine at 0xaaaaf37c7710
Caught exception: EVP_ENGINE_by_id(pkcs11) failed: - SSL error:1300006D:engine routines::init failed
- SSL error:13000074:engine routines::no such engine
```

With finish and free, without remove:
```
Reloading engine.
calling ENGINE_finish(0xaaaafef42710)
calling ENGINE_free(0xaaaafef42710)
Info: freed 'pkcs11' engine at 0xaaaafef42710
destroyPkcs11Object: trying to delete 1 objects by keyid, class 2
destroyPkcs11Object: trying to delete 1 objects by keyid, class 3
Private and public key 'testkey' have been deleted by id (again).
Private and public key 'testkey' have been generated (again).
Trying to check key validity by signing and verifying (again).
Public key:
-----BEGIN PUBLIC KEY-----
MHYwEAYHKoZIzj0CAQYFK4EEACIDYgAEzVdaE85BmSK0TsQyjNociB+VbfiV9zA6
uo2eVd8CF15GR8Ce/uaVuqNpsIXmhbKUYlpp1CJV8iX17ZPnjmNhs90Oy3jWN3YM
5Ds5H/olnu9CdeuoV+OowKLavWZ9NmoR
-----END PUBLIC KEY-----

Signature created
Verification failed
```

What am I doing differently?

sbrandt-aixtrusion:
Simplification:

```cpp
    ENGINE_load_builtin_engines();
    int ret;
    {
        auto eng = ENGINE_by_id("pkcs11");
        printf("ENGINE_by_id('pkcs11') -> %p\n", (void*)eng);
        ret = ENGINE_init(eng);
        printf("ENGINE_init(%p) -> %d\n", (void*)eng, ret);

        ret = ENGINE_free(eng);
        printf("ENGINE_free(%p) -> %d\n", (void*)eng, ret);
        ret = ENGINE_remove(eng);
        printf("ENGINE_remove(%p) -> %d\n", (void*)eng, ret);
    }
    {
        auto eng = ENGINE_by_id("pkcs11");
        printf("ENGINE_by_id('pkcs11') -> %p\n", (void*)eng);
    }
```
results in 
```
ENGINE_by_id('pkcs11') -> 0xaaab0d029aa0
ENGINE_init(0xaaab0d029aa0) -> 1
ENGINE_free(0xaaab0d029aa0) -> 1
ENGINE_remove(0xaaab0d029aa0) -> 1
ENGINE_by_id('pkcs11') -> (nil)
```
which means the engine cannot be loaded again.

If I remove the call to ENGINE_init, I can reload it after remove:
```
ENGINE_by_id('pkcs11') -> 0xaaab0c0c2aa0
ENGINE_free(0xaaab0c0c2aa0) -> 1
ENGINE_remove(0xaaab0c0c2aa0) -> 1
ENGINE_by_id('pkcs11') -> 0xaaab0c0c2aa0
```
but that does not really help me or anyone.

sbrandt-aixtrusion:
Okay, it does work if ENGINE_finish() is also called .. but just in the simple version for now.

mtrojnar:
Can you please open an issue for the second `ENGINE_by_id('pkcs11')` failing?  This is not something that could possibly be addressed by this PR.

sbrandt-aixtrusion:
See #556

Issue #551, opened 2024-10-22
Build a static pkcs11 engine library


maroueneboubakri:
Hello,

Has anyone succeeded to build a static pkcs11 library ? Despite that engine are designed to be dynamically loaded, but programmatically we still can set and register an engine in my application which would be statically linked against OpenSSL library and PKCS11 engine and would be great if the build system supports it, or maybe it is supported, but how to generate it.

Best
Maro

maroueneboubakri:
@mtrojnar FYI

maroueneboubakri:
With `--disable-shared  --enable-static ` pkcs11 library is not generate for Linux host, while for MinGW it is generating a static pkcs11.a library.

mtrojnar:
How do you intend to use a static OpenSSL engine?

maroueneboubakri:
> How do you intend to use a static OpenSSL engine?

Using `ENGINE_add()` to add a new engine to OpenSSL's internal list. The engine front has `bind_helper()` that can be exported an used to create the engine.

Best
Maro

mtrojnar:
I admit I have zero experience with using static engines with OpenSSL.  Feel free to submit a PR to add support for that.

PR #560, opened 2024-11-19
src/Makefile.am: Build static PKCS#11 engine library (libpkcs11.a)

maroueneboubakri:
As discussed in issue #551, an application may statically link against the OpenSSL library. Consequently, using a dynamically loadable engine may present challenges. This PR proposes modifying the build system to enable the creation of a static PKCS11 engine library instead (libpkcs11.a).

The library exposes the `bind_engine()` function, which allows the application to load and register the engine directly within the application:

	extern int bind_engine(ENGINE *e, const char *id);

Below is a sample code snippet demonstrating how to load and register the engine:

	ENGINE *e = NULL;
	e = ENGINE_new();
	if (e == NULL) {
		error("Error creating new engine instance: %s\n",
		      ERR_reason_error_string(ERR_get_error()));
		return 0;
	}

	/* Bind the engine using the bind_engine function */
	if (!bind_engine(e, "pkcs11")) {
		error("Error binding pkcs11 engine: %s\n",
		      ERR_reason_error_string(ERR_get_error()));
		ENGINE_free(e);
		return 0;
	}

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -6,7 +6,7 @@ EXTRA_DIST = Makefile.mak libp11.rc.in pkcs11.rc.in
 
 noinst_HEADERS= libp11-int.h pkcs11.h p11_pthread.h
 include_HEADERS= libp11.h p11_err.h
-lib_LTLIBRARIES = libp11.la
+lib_LTLIBRARIES = libp11.la libpkcs11.la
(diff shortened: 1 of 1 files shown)


maroueneboubakri:
@mtrojnar FYI.

PR #562, opened 2024-11-21
Improved tests

olszomal:
- Centralized the use of the PKG_CONFIG_PATH environment variable to determine the OpenSSL version required by libp11.
- Added flexibility to configure the OpenSSL binary via LD_LIBRARY_PATH for custom shared library locations.
- Forced the use of the locally built engine by setting OPENSSL_ENGINES to ../src/.libs/.
- Tests rely on pre-installed pkcs11-tool and softhsm2-util, which must be accessible through the default system PATH.
- Added a guide for debugging the engine with Valgrind.
- Refactored the code to improve clarity, simplify logic, and eliminate redundancies.
- Switched from sh to bash to leverage Bash-specific features and ensure compatibility.
- Removed legacy support checks for OpenSSL versions older than 1.0.0.
- Skipped tests when libsofthsm2.so is built with libcrypto.so.3, as it is incompatible with OpenSSL 1.0.2u.
- Update copyright information to Mobi - Com Polska Sp. z o.o.


mtrojnar:
Please update copyright for those changes from Michał Trojnara <Michal.Trojnara@stunnel.org> to Mobi - Com Polska Sp. z o.o.

PR #585, opened 2025-03-14
libp11 PKCS#11 provider support

olszomal:
This code was sponsored by Mobi-Com Polska Sp. z o.o.


dengert:
Note:  RedHat also has a OpenSSL 3 provider:  https://github.com/latchset/pkcs11-provider

And library names and other names may be  the same between the latchset and libp11 versions. 

I would suggest that both both  latchset and libp11 address the issue

mtrojnar:
@simo5 These are resources I paid for. I'm not telling you how to spend your money, and I ask that you not tell me how to spend mine.

simo5:
> @simo5 These are resources I paid for. I'm not telling you how to spend your money, and I ask that you not tell me how to spend mine.

You are free to burn money in a fire for what I care, I do not intend to interact further.

dengert:
Not exactly the response I was expecting.

As an OpenSC developer, I expect to see bug reports when trying to use either of these providers when using `opensc-pkcs11.so`

I was hoping that the name of the provider would give a clue to where it the provider's package.
It look like latchset's name is `pkcs11-provider.so`  and libp11's name is `pkcs11prov.so`.  And both use `[pkcs11_sect]` in openssl.conf.  But I would not expect both be loaded at the same time, although that might be possible.

PR #589, opened 2025-04-07
Squash Coverity false positives for out-of-bounds

olszomal:


mtrojnar:
Did you mean: *[KEY-ID] is **not** a null-terminated string*?

PR #603, opened 2025-06-09
Allow OS-based thread synchronization in `C_Initialize()` call

### LINKED ISSUE #602: Use of OS locking disabled in 0.4.14
The [NEWS and copyright updates](https://github.com/OpenSC/libp11/commit/a2b4f8a095dbf601e224617f00601327357a7f4b) commit changes the call to `C_Initialize()` in `src/p11_load.c` to not set the `CKF_OS_LOCKING_OK` flag.

This breaks (at least) the `p11-kit` PKCS#11 RPC client on Alpine 3.21.3, which (apparently) relies on OS locking.

Not sure, but it looks like this change may have been inadvertent, based on a) the lack of any reference to locking in the commit description, and b) the similarity to the fix in [this comment](https://github.com/OpenSC/libp11/issues/595#issuecomment-2873485031) to #595, which suggests the change might have been done while investigating that issue and later accidentally included in the commit.

nojocodex:
Fixes #602.

### DIFF
--- a/src/p11_load.c
+++ b/src/p11_load.c
@@ -77,7 +77,7 @@ static int pkcs11_initialize(PKCS11_CTX_private *cpriv)
 
 	memset(&args, 0, sizeof(args));
 	/* Unconditionally say using OS locking primitives is OK */
-	// args.flags |= CKF_OS_LOCKING_OK;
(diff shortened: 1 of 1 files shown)


mtrojnar:
My bad.  I committed a change that was only intended for internal testing.

PR #657, opened 2026-06-11
Add opt-in no_login_cache to release the token login session on key free.

dantecmelo:
### Pull Request Type

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### Related Issue
Issue number: N/A

### Current Behavior

libp11 caches the token login session for the lifetime of the process and would
normally release it at exit. However, its exit-time teardown is deliberately
skipped: the `atexit` handler in `util_uri.c` sets `g_shutdown_mode`, which makes
`UTIL_CTX_free_libp11` skip `PKCS11_release_all_slots` / `C_Finalize` (to avoid
calling into the module during OpenSSL's teardown). As a result the login session
is never released.

On a token with a hard limit on concurrent authenticated sessions — notably the
YubiHSM 2 (max 16) — running one short-lived `openssl` process per file leaks one
session per process (on the YubiHSM, `C_CloseSession` does not free the
authenticated session; only `C_Logout` / `C_Finalize` does). After 16 files the
17th fails with `CKR_SESSION_COUNT` ("could not read private key").

### New Behavior

Adds an opt-in `no_login_cache` provider parameter (env `PKCS11_NO_LOGIN_CACHE`)
and a matching public `PKCS11_set_no_login_cache()` API. When enabled, the token
is logged out and its sessions are closed as soon as a private key object is
freed — during normal runtime instead of at the skipped exit. Re-login happens
automatically on the next key load. With the parameter unset, behaviour is
byte-for-byte unchanged.

### Scope of Changes

- Core: a self-contained, session-only logout helper
  `pkcs11_slot_logout_session_only()` (`p11_slot.c`), a per-context/per-slot
  `no_login_cache` flag, and the public `PKCS11_set_no_login_cache()`
  (`p11_front.c`, `libp11.h`, `libp11.exports`).
- Provider front-end: a `no_login_cache` parameter mirroring `force_login`
  (`provider_helpers.c`, `util_uri.c`, `util.h`). Because the provider's
  `keymgmt_load` transfers ownership of the key to OpenSSL, the wrapped key's
  legacy RSA/EC `finish` callback does not run during the process — so the logout
  is triggered from `keymgmt_free` (`provider.c`) via `p11_keydata_release_login()`,
  which reaches the slot through the wrapped key's RSA/EC ex-data
  (`pkcs11_release_login_for_pkey()` in `p11_key.c`).
- Engine front-end: the same release is triggered from the RSA/EC key `finish`
  callbacks (`p11_rsa.c`, `p11_ec.c`), where keys are freed via that path.

### Testing

- [ ] Existing tests
- [ ] New tests added
- [x] Manual testing

Verified on a YubiHSM 2 (max 16 sessions) with the OpenSSL 3 provider on Debian:
a 50-iteration loop of `openssl dgst -sha512 -sign`, one process per file,
previously failed at the 17th with `CKR_SESSION_COUNT`; with `no_login_cache = 1`
it now completes all 50. A PKCS#11 trace shows one `C_Logout` per key free, and
the device's active session count returns to baseline between runs.

### Additional Notes

- Opt-in and backward compatible: with the parameter unset, behaviour is
  byte-for-byte unchanged.
- Safety / re-entrancy: the logout helper is self-contained — it does not call
  `pkcs11_wipe_cache` or `pkcs11_get_session`, so it cannot re-enter the key-free
  callback or nest `slot->lock`; it is idempotent via a lockless `logged_in` check.
- Key-type scope: covers RSA and ECDSA (EC). EdDSA (Ed25519/Ed448) is out of
  scope — libp11 has no per-key finish callback for it, so there is no equivalent
  hook.
- Trade-off: intended for many short-lived signing processes. A long-lived
  process that signs repeatedly benefits from the cached login, so it should
  leave the option off — hence opt-in.
- Background: the skipped exit-time teardown that makes runtime release necessary
  is the behaviour libp11 adopted to avoid an exit-time crash/deadlock (cf. #527).

## License Declaration


olszomal:
Could you please try to reproduce the issue on current master with `PKCS11_FORCE_CLEANUP=1` set?

mtrojnar:
The author does not respond.  I'm closing this PR as stale.

PR #658, opened 2026-06-16
Add PQC signature support

### LINKED ISSUE #606: Support for Vendor Defined Key Types and mechanisms
Is there any plan to add support to vendor defined key types and mechanisms? Right now the openssl engine rejects any key that is not RSA or EC. Is this really necessary? You could simply pass plainly the CKK_VENDOR_DEFINED specific value to C_Sign (for example).

Maybe this should be more flexible looking for future grand modifications in the cryptography landscape like the one we are living with PQC.

olszomal:
### Pull Request Type

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### Related Issue

Fixes #606

### Current Behavior

The PKCS#11 provider supports traditional key types and `EdDSA`, but it does not handle post-quantum signature algorithms.

### New Behavior

`ML-DSA`, `SLH-DSA`, and `FALCON` keys can be detected, generated, signed with, and verified where supported by the token and OpenSSL version. 

### Scope of Changes

- Added PKCS#11 provider support for `ML-DSA`, `SLH-DSA`, and `FALCON` signing and verification.
- Added PQC key pair generation.
- Added PQC examples and provider tests.
- Fixed PKCS#11 session handling and cleanup paths.
- Fixed attribute template cleanup to avoid leaks.
- Improved compatibility with newer OpenSSL APIs.
- Fixed MSVC compiler warnings.
- Updated documentation, build files, and tests.

### Testing

- [x] Existing tests
- [x] New tests added
- [x] Manual testing -> performed with Thales TCT Luna T-5000 Network HSM

### Additional Notes

- OpenSSL 3.5 adds native support for the PQC algorithms `ML-DSA` and `SLH-DSA`. 
- `FALCON` is not supported by OpenSSL, so public-key operations for `FALCON` are performed through the PKCS#11 token instead of relying on the OpenSSL provider implementation.
- `FALCON` and `SLH-DSA` support could not be tested because no token supporting these mechanisms was available during development.

## License Declaration


mtrojnar:
Thanks for PR. I found few issues worth fixing before merge:

1. **ML-DSA signing fails with SoftHSM**
   - `src/p11_pkey.c:869`
   - `CKM_ML_DSA` signing uses `pParameter = NULL`. SoftHSM ML-DSA expects `CK_SIGN_ADDITIONAL_CONTEXT` even for empty context.
   - New test fails:
     - `make -C tests check TESTS='provider-mldsa87-keygen.softhsm'`
     - failure: `EVP_Digest_sign_verify_test() failed with err code: -6`
   - `pkcs11-spy` shows `C_Sign` returns `CKR_GENERAL_ERROR`.
   - `pkcs11-tool` signs same key successfully because it passes `CKH_HEDGE_PREFERRED`, `pContext = NULL`, `ulContextLen = 0`.
   - Suggested fix: define/pass `CK_SIGN_ADDITIONAL_CONTEXT` for ML-DSA signing.

2. **SLH-DSA mechanism constants are off by one**
   - `src/pkcs11.h:537`
   - PR defines:
     - `CKM_SLH_DSA_KEY_PAIR_GEN = 0x2E`
     - `CKM_SLH_DSA = 0x2F`
   - PKCS#11 3.2 / p11-kit define:
     - `CKM_SLH_DSA_KEY_PAIR_GEN = 0x2D`
     - `CKM_SLH_DSA = 0x2E`
   - Compliant tokens will reject or run wrong mechanism. Please adjust constants.

3. **Public API symbol declared but not exported**
   - `src/libp11.h:568`
   - `PKCS11_evp_pkey_verify()` is declared public, but missing from `src/libp11.exports`.
   - External users would compile but fail linking. Either export it and review ABI/versioning, or keep it internal.

4. **New ML-DSA DER fixtures missing from dist data**
   - `tests/Makefile.am:81`
   - Added files `mldsa87-cert.der`, `mldsa87-privkey.der`, `mldsa87-pubkey.der` are not in `dist_check_DATA`.
   - `make distcheck` / release tarballs will miss them. Please add them.

Also `git diff --check` reports trailing whitespace:
- `src/p11_pkey.c:221`

olszomal:
Thanks a lot for the review.

Regarding point 1, I checked the `ML-DSA` signing failure with SoftHSM2. This appears to be caused by the current SoftHSM2 master issue fixed by SoftHSMv2 PR softhsm/SoftHSMv2#881

That change explicitly uses "provider=default" when creating internal `ML-DSA` key contexts. With that fix applied, `ML-DSA` signing works correctly.

The same `ML-DSA` signing path also works with a Thales TCT Luna T-5000 Network HSM.

I fixed the remaining issues:
- corrected the `SLH-DSA` mechanism constant values,
- added `PKCS11_evp_pkey_verify` to libp11.exports,
- added the new `ML-DSA` DER fixtures to `dist_check_DATA`,
- fixed the trailing whitespace reported by `git diff --check`.

mtrojnar:
I understand you had no means for testing the FALCON code yet, but addressing the following issues now may speed up things once a testing platform is available:

1. `src/p11_falcon.c:265` forces FALCON public-key creation through `provider=default`. Default provider does not implement FALCON, while this PR advertises FALCON keymgmt from `pkcs11prov` (`src/provider.c:238`).
Result: `EVP_PKEY_new_raw_public_key_ex()` returns NULL for FALCON keys, so FALCON load/sign/verify never reaches PKCS#11.
Fix: create FALCON public keys with `pkcs11prov` / matching property query, or omit default-only propq.

3. `src/provider_helpers.c:2847` detects FALCON with `EVP_PKEY_is_a(pkey, "FALCON512")` / `"FALCON1024"`, but provider advertises and creates `"FALCON-512"` / `"FALCON-1024"` (`src/provider.c:238`, `src/p11_falcon.c:272`).
Result: even if FALCON key creation succeeds, `evp_pkey_get_type_id()` can return `EVP_PKEY_NONE`, leaving key metadata unset and operations failing.
Fix: use same names everywhere, preferably include both hyphenated and non-hyphen aliases if needed.

5. `src/p11_key.c:1175` attaches PKCS#11 ex_data only to private EVP_PKEYs. FALCON verify path uses public EVP_PKEY (`src/provider.c:1432`), then `PKCS11_evp_pkey_verify()` calls `pkcs11_get_pkcs11_key()` and requires that ex_data (`src/p11_front.c:875`).
Result: public-key FALCON verify always fails before `C_VerifyInit`.
Fix: preserve public PKCS#11 object reference in provider keydata or attach ex_data for public FALCON keys too, with correct lifetime/refcount handling.

olszomal:
I addressed the FALCON issues:
- FALCON public keys now use the pkcs11prov keymgmt path instead of `provider=default`.
- FALCON naming is now consistent with the advertised keymgmt names, with aliases added for `FALCON-512`, `FALCON-1024`, `falcon512`, `falcon1024`, `FN-DSA-512`, and `FN-DSA-1024`.
- Public EVP_PKEYs now keep the backing PKCS#11 object in `ex_data`, so token-side FALCON verify can reach `C_VerifyInit()`.

Thanks for catching these.

mtrojnar:
I found a few issues worth investigating:

- [High] [`src/p11_falcon.c:36`](https://github.com/olszomal/libp11/blob/pqc_signatures/src/p11_falcon.c#L36) Falcon-1024 public key length is `2305`, which is secret-key length. Public key length should be `1793`. Valid Falcon-1024 public keys will be rejected. Also [`src/provider_helpers.c:77`](https://github.com/olszomal/libp11/blob/pqc_signatures/src/provider_helpers.c#L77) uses padded Falcon signature sizes (`666`/`1280`) while names/OIDs say Falcon, so signing may under-allocate unless padded variants are intended.

- [Medium] [`src/p11_mldsa.c:221`](https://github.com/olszomal/libp11/blob/pqc_signatures/src/p11_mldsa.c#L221), [`src/p11_slhdsa.c:221`](https://github.com/olszomal/libp11/blob/pqc_signatures/src/p11_slhdsa.c#L221), and [`src/p11_falcon.c:220`](https://github.com/olszomal/libp11/blob/pqc_signatures/src/p11_falcon.c#L220) acquire sessions but never return them. Repeated PQC key loads can exhaust token sessions. Add `pkcs11_put_session(slot, session)` on all exit paths, like EdDSA does.

- [Medium] [`src/provider.c:828`](https://github.com/olszomal/libp11/blob/pqc_signatures/src/provider.c#L828) routes direct `EVP_PKEY_verify()` through default-provider verification. Falcon has no default-provider verify path, while one-shot digest verify has token-side special handling at [`src/provider.c:1398`](https://github.com/olszomal/libp11/blob/pqc_signatures/src/provider.c#L1398). Direct verify needs same Falcon token-side path.

- [Medium] [`src/libp11.h:568`](https://github.com/olszomal/libp11/blob/pqc_signatures/src/libp11.h#L568) adds exported `PKCS11_evp_pkey_verify`, and [`src/libp11.h:575`](https://github.com/olszomal/libp11/blob/pqc_signatures/src/libp11.h#L575) changes public `PKCS11_evp_pkey_decrypt` argument type, but [`configure.ac:83`](https://github.com/olszomal/libp11/blob/pqc_signatures/configure.ac#L83) libtool version is unchanged. ABI/version bump or review needed.

olszomal:
Thanks for the review. I addressed the issues:
- fixed the Falcon-1024 public key length,
- added missing `pkcs11_put_session()` calls in the PQC public-key load paths,
- routed direct Falcon `EVP_PKEY_verify()` through the token-side verify path,
- updated `configure.ac` libtool versioning for the ABI/API changes.

Please pay particular attention to the `configure.ac` changes and confirm whether this ABI/libtool bump is the correct approach here.

olszomal:
I reset the `configure.ac` libtool-versioning changes.

mtrojnar:
## Findings

 ### [Medium] src/p11_key.c:303 — Mechanism constant used as key type for Falcon

 In pkcs11_object_from_handle, the switch (key_type) block matches CKM_PQC_FALCON as a key type. But CKA_KEY_TYPE holds a CKK_ key-type constant, not a CKM_ mechanism constant. CKM_PQC_FALCON (CKM_VENDOR_DEFINED + 0x10025UL) is a mechanism, not a key type. This works only if the token happens to write this exact
 value into CKA_KEY_TYPE — vendor-specific and fragile. ML-DSA/SLH-DSA correctly use CKK_ML_DSA / CKK_SLH_DSA.

 Suggested fix: Define a CKK_FALCON key-type constant or use CKK_VENDOR_DEFINED and disambiguate via CKA_PARAMETER_SET.

 ### [Low] src/libp11.h:578 — ABI break: int → size_t on oaep_labellen

 PKCS11_evp_pkey_decrypt changes the oaep_labellen parameter from int to size_t. On LP64 platforms this changes parameter size from 4 to 8 bytes. Code compiled against the old header will pass/read wrong stack layout when calling a new libp11 binary. Guarded by #if OPENSSL_VERSION_NUMBER >= 0x30000000L, limiting
 blast radius but still a public-ABI change.

 ### [Low] src/provider_helpers.c:~490 — Duplicated data_type check in p11_public_equal

 The g1->data_type != OSSL_PARAM_UTF8_STRING check appears twice in the same if (g1 != NULL && g2 != NULL) block — once before the string extraction, once after the name comparison. Second occurrence is unreachable dead code. Harmless but confusing.

 ### [Low] src/p11_key.c:432 — Redundant pkcs11_zap_attrs after pkcs11_object_from_template

 pkcs11_object_from_object calls pkcs11_zap_attrs(&tmpl) after pkcs11_object_from_template, but pkcs11_handle_from_template (called inside) already zaps the template. The second zap hits a zeroed template (no-op), so harmless. Confusing to maintainers.

 ### [Info] src/p11_rsa.c:211 — Pre-existing session leak on RSA_new() OOM

 The refactor moved pkcs11_put_session to after RSA_new() inside pkcs11_get_evp_key_rsa. If RSA_new() returns NULL, the session is leaked (both old and new code have this pre-existing leak on failure paths). The pkcs11_private_encrypt/pkcs11_private_decrypt session leak fixes are correct.

 ### [Info] src/p11_load.c:27-28 — Falcon OIDs use enterprise arc 1.3.9999

 Falcon OIDs are registered as 1.3.9999.3.11 / 1.3.9999.3.14. These appear to be placeholder/test OIDs, not the official NIST/IANA-assigned Falcon OIDs. Interop risk with other Falcon implementations using different OIDs.

 ### Positive fixes

 - Session leaks plugged: p11_rsa.c (pkcs11_private_encrypt/_decrypt fork-check path), p11_eddsa.c (pkcs11_get_raw_public_key error label, pkcs11_eddsa_pmeth_sign/_digestsign), p11_key.c (pkcs11_ec_keygen early returns, pkcs11_eddsa_keygen unsupported-nid path).
 - EVP_PKEY_is_a migration in p11_key.c and provider_helpers.c improves OpenSSL 4.x forward-compat.
 - p11_public_equal rewrite to parameter-based key comparison is more robust than type-enum dispatch.
 - is_oneshot_sig_type correctly reports OSSL_PKEY_PARAM_MANDATORY_DIGEST = "UNDEF" for one-shot algorithms (EdDSA, ML-DSA, SLH-DSA, Falcon).
 - PKCS11_evp_pkey_verify new public function enables provider-side Falcon token verification.
 - oaep_labellen overflow check added: if (oaep_labellen > (size_t)((CK_ULONG)-1)) return -1;.

## Tests

 Tests cannot be run without SoftHSM2 built with PQC support (requires OpenSSL 3.5+ with ML-DSA/SLH-DSA enabled and a SoftHSM that supports these mechanisms). The PR adds test fixtures (tests/mldsa87-*.der, tests/provider-mldsa87-*.softhsm) and test binaries (tests/mldsa87-keygen-prov.c). No test regressions in
 existing non-PQC tests expected from the diffs (the changed code paths for RSA/EC/EdDSA are minimal and appear correct).

## Notes

 - Falcon keymgmt entries in provider.c (p11_keymgmts[]) are unconditionally compiled while ML-DSA/SLH-DSA entries have #if OPENSSL_VERSION_NUMBER >= 0x30500000L. OpenSSL should ignore unknown algorithm names, but consistency would be better.
 - Build tested only via git diff analysis; full ./configure --enable-strict && make not attempted due to OpenSSL version constraints for PQC algorithms.
 - The PKCS11_EDDSA_KGEN → PKCS11_NID_KGEN rename with #define backward compat is clean.

olszomal:
Thanks for the review. I fixed the reported issues and documented the provisional OQS Falcon OIDs and Luna-specific Falcon constants.

PR #662, opened 2026-07-22
Add missing `p11_ver.h` file to installed headers

botantony:
### Pull Request Type

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### Related Issue

Issue number: N/A

### Current Behavior

`make install` does not install `p11_ver.h` header used by `p11_err.h`

### New Behavior

`p11_ver.h` is installed to `includes`

### Testing

- [ ] Existing tests
- [ ] New tests added
- [x] Manual testing (tested locally, `make install` works as expected)

### Additional Notes

Found in https://github.com/Homebrew/homebrew-core/pull/294427

## License Declaration

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -9,7 +9,7 @@ EXTRA_DIST = Makefile.mak libp11.rc.in pkcs11.rc.in
 
 # Headers
 noinst_HEADERS= libp11-int.h pkcs11.h p11_pthread.h provider_helpers.h util.h
-include_HEADERS= libp11.h p11_err.h
+include_HEADERS= libp11.h p11_err.h p11_ver.h
 
 lib_LTLIBRARIES = libp11.la
 enginesexec_LTLIBRARIES =


mtrojnar:
"`make install` does not install `p11_ver.h` header used by `p11_err.h`" suggests that the distribution tarball is unbuildable.  We shouldn't have to guess the claimed error symptoms though.

Is it possible that we never executed `make distcheck` after adding new source files?  If this is the case, such repeating pattern is getting annoying.  We may need `make distcheck` in our GitHub Actions CI.

olszomal:
This patch installs `p11_ver.h` as a public header.

Without it, compiling an external application such as:
```c
#include <libp11.h>

int main(void)
{
	return 0;
}
```
against the installed headers fails with:
```
fatal error: p11_ver.h: No such file or directory
```
`make install` and `make distcheck` do not catch this because they do not test an external build using only the installed headers.

LGTM, approved.

